PR #29, opened 2023-07-14
CTX-4247: Created a bioinformatics folder, moved the qiime2 folder th…

VukManojlovic:
…ere and migrated shared code from templates related to sequence alignment to sequence_alignment

### DIFF
--- a/coretex/bioinformatics/qiime2/__init__.py
+++ b/coretex/bioinformatics/qiime2/__init__.py
@@ -21,24 +21,14 @@ from pathlib import Path
 import subprocess
 import logging
 
+from ..utils import logProcessOutput
+
 
 class QiimeCommandException(Exception):
     pass
 
 
-def logProcessOutput(output: bytes, severity: int) -> None:
-    decoded = output.decode("UTF-8")
-
-    for line in decoded.split("\n"):
-        # skip empty lines
-        if line.strip() == "":
-            continue
-
-        # ignoring type for now, has to be fixed in coretexpylib
-        logging.getLogger("coretexpylib").log(severity, line)
-
-
-def QiimeCommand(args: List[str]) -> None:
+def command(args: List[str]) -> None:
     process = subprocess.Popen(
         args,
         shell = False,
@@ -82,7 +72,7 @@ def toolsImport(
             "--input-format" , inputFormat
         ])
 
-    QiimeCommand(args)
+    command(args)
 
 
 def demuxEmpSingle(
@@ -93,7 +83,7 @@ def demuxEmpSingle(
     errorCorretctionDetailsPath: str
 ) -> None:
 
-    QiimeCommand([
+    command([
         "qiime", "demux", "emp-single",
         "--i-seqs", sequencesPath,
         "--m-barcodes-file", barcodesPath,
@@ -105,7 +95,7 @@ def demuxEmpSingle(
 
 
 def demuxSummarize(dataPath: str, visualizationPath: str) -> None:
-    QiimeCommand([
+    command([
         "qiime", "demux", "summarize",
         "--i-data", dataPath,
         "--o-visualization", visualizationPath
@@ -121,7 +111,7 @@ def dada2DenoiseSingle(
     denoisingStatsPath: str
 ) -> None:
 
-    QiimeCommand([
+    command([
         "qiime", "dada2", "denoise-single",
         "--i-demultiplexed-seqs", inputPath,
         "--p-trim-left", str(trimLeft),
@@ -133,7 +123,7 @@ def dada2DenoiseSingle(
 
 
 def metadataTabulate(inputFile: str, visualizationPath: str) -> None:
-    QiimeCommand([
+    command([
         "qiime", "metadata", "tabulate",
         "--m-input-file", inputFile,
         "--o-visualization", visualizationPath
@@ -141,7 +131,7 @@ def metadataTabulate(inputFile: str, visualizationPath: str) -> None:
 
 
 def featureTableSummarize(inputPath: str, visualizationPath: str, metadataPath: str) -> None:
-    QiimeCommand([
+    command([
         "qiime", "feature-table", "summarize",
         "--i-table", inputPath,
         "--o-visualization", visualizationPath,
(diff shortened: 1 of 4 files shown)


coretex/bioinformatics/sequence_alignment/__init__.py
```diff
@@ -0,0 +1,132 @@
...
+            if process.returncode == 0:
+                logProcessOutput(stderr, logging.INFO)
+            else:
+                logProcessOutput(stderr, logging.CRITICAL)
+
+    if process.returncode != 0:
+        raise RuntimeError(f">> [Coretex] Falied to execute command. Returncode: {process.returncode}")
+
+
+def indexCommand(bwaPath: Path, sequencePath: Path, prefix: Path) -> None:
```
igorperic17, line 57: Let's try adding some user-friendly comments for each command line argument here, so the usage of third party tools is a bit more obvious.

coretex/bioinformatics/utils.py
```diff
@@ -0,0 +1,30 @@
...
+#     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#     GNU Affero General Public License for more details.
+
+#     You should have received a copy of the GNU Affero General Public License
+#     along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+import logging
+
+
+def logProcessOutput(output: bytes, severity: int) -> None:
```
igorperic17, line 21: Are we sure this will not be a memory (overflow) problem with long logs?

coretex/bioinformatics/sequence_alignment/__init__.py
```diff
@@ -0,0 +1,132 @@
...
+        str(bwaPath.absolute()), "mem",
+        "-o", str(outputPath.absolute()),
+        str(prefix.absolute()),
+        str(sequencePath.absolute())
+        ]
+
+    command(args, True)
+
+
+def sam2bamCommand(samtoolsPath: Path, samPath: Path, outputPath: Path) -> None:
```
igorperic17, line 76: Please add doc strings to each command, including URLs to the tools used or relevant documentation.

coretex/bioinformatics/sequence_alignment/__init__.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+def extractData(samtoolsPath: Path, file: Path) -> Tuple[List[int], List[int], List[int]]:
+    scores: list[int] = []
+    positions: list[int] = []
+    sequenceLengths: list[int] = []
+
+    args = [
+        str(samtoolsPath.absolute()),
+        "view", "-F", "256", "-F", "2048", "-F", "512",
```
igorperic17, line 92: What are these parameters? Should we make them experiment params perhaps and let the user change them for their experiment?

coretex/bioinformatics/qiime2/__init__.py
```diff
@@ -21,24 +21,14 @@
 import subprocess
 import logging
 
+from ..utils import logProcessOutput
+
 
 class QiimeCommandException(Exception):
```
dule1322, line 27: let's rename this too since the `QiimeCommand` has been renamed too

coretex/bioinformatics/sequence_alignment/__init__.py
```diff
@@ -0,0 +1,132 @@
...
+#     along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+from typing import List, Tuple
+from pathlib import Path
+
+import subprocess
+import logging
+
+from ...coretex import CustomDataset
+from ..utils import logProcessOutput
```
dule1322, line 25: imports not sorted properly

coretex/bioinformatics/sequence_alignment/__init__.py
```diff
@@ -0,0 +1,132 @@
...
+            logProcessOutput(stdout, logging.INFO)
+
+        if len(stderr) > 0 and not ignoreOutput:
+            if process.returncode == 0:
+                logProcessOutput(stderr, logging.INFO)
+            else:
+                logProcessOutput(stderr, logging.CRITICAL)
+
+    if process.returncode != 0:
+        raise RuntimeError(f">> [Coretex] Falied to execute command. Returncode: {process.returncode}")
```
dule1322, line 54: maybe it should be printed out what command failed

coretex/bioinformatics/sequence_alignment/__init__.py
```diff
@@ -0,0 +1,132 @@
...
+    ])
+
+
+def alignCommand(bwaPath: Path, prefix: Path, sequencePath: Path, outputPath: Path) -> None:
+    args = [
+        str(bwaPath.absolute()), "mem",
+        "-o", str(outputPath.absolute()),
+        str(prefix.absolute()),
+        str(sequencePath.absolute())
+        ]
```
dule1322, line 71: formatting

coretex/bioinformatics/sequence_alignment/__init__.py
```diff
@@ -0,0 +1,132 @@
...
+    positions: list[int] = []
+    sequenceLengths: list[int] = []
+
+    args = [
+        str(samtoolsPath.absolute()),
+        "view", "-F", "256", "-F", "2048", "-F", "512",
+        str(file.absolute())
+    ]
+
+    process = subprocess.Popen(
```
dule1322, line 96: why is `command` function not used here for executing the command?

coretex/bioinformatics/sequence_alignment/__init__.py
```diff
@@ -0,0 +1,132 @@
...
+from pathlib import Path
+
+import subprocess
+import logging
+
+from ...coretex import CustomDataset
+from ..utils import logProcessOutput
+
+
+def command(args: List[str], ignoreOutput: bool = False) -> None:
```
dule1322, line 28: why do we have duplicate `command` function?

coretex/bioinformatics/sequence_alignment/__init__.py
VukManojlovic: These parameters are used to filter out all duplicate reads when reading the .bam file (three -F ### for three types of duplicates). So it would be unnecessary to make these experiment parameters.

coretex/bioinformatics/sequence_alignment/__init__.py
VukManojlovic: The samtools view command outputs it's results to stdout, so in order to capture that output we use subprocess directly in the extractData function to write all relevant information into lists. The command function would simply print the stdout as it is a generic function.

dule1322: Understood. Can we also catch `stderr` output here since if something goes wrong there might be some useful info there?

PR #34, opened 2023-07-24
CTX-4356: Added cache name hashing to utils

VukManojlovic:


coretex/utils/hash.py
```diff
@@ -0,0 +1,16 @@
+import base64
+import hashlib
+
+
+maxNameLength = 50
+
+
+def md5(object: bytes) -> bytes:
+    return hashlib.md5(object).digest()
```
dule1322, line 9: there is no need for this to be a separate function

PR #40, opened 2023-08-02
CTX-4421: Added cutadaptTrim command

VukManojlovic:


coretex/bioinformatics/__init__.py
```diff
@@ -15,4 +15,35 @@
 #     You should have received a copy of the GNU Affero General Public License
 #     along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-from .utils import CommandException
+from typing import Optional
+
+from .utils import CommandException, command
+
+
+def cutadaptTrim(
```
dule1322, line 23: function is missing documentation

coretex/bioinformatics/__init__.py
```diff
@@ -15,4 +15,35 @@
...
+    forewardOutput: str,
+    forewardAdapter: str,
+    reverseFile: Optional[str] = None,
+    reverseOutput: Optional[str] = None,
+    reverseAdapter: Optional[str] = None
+) -> None:
+
+    args: list[str] = [
+        "cutadapt",
+        "-o", forewardOutput,
```
dule1322, line 34: forward is misspelled

coretex/bioinformatics/__init__.py
```diff
@@ -15,4 +15,35 @@
...
 #     along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-from .utils import CommandException
+from typing import Optional
+
+from .utils import CommandException, command
+
+
+def cutadaptTrim(
+    forewardFile: str,
```
dule1322, line 24: why are we using `str` instead of `Path` for paths?

VukManojlovic: Requiring strings is the convention that all other command wrapper functions in the bioinformatics subdirectory use, so this one should not be an exception

dule1322: For now I will resolve this but this needs to be changed

PR #41, opened 2023-08-04
CTX-4421: Added documentation to the new functions and added compatability with Path object for them

VukManojlovic:
Added documentation and Path object support to :
- dada2DenoisePaired in bioinformatics/ctx_qiime2
- cutadaptTrim in bioinformatics

### DIFF
--- a/coretex/bioinformatics/__init__.py
+++ b/coretex/bioinformatics/__init__.py
@@ -19,6 +19,7 @@ from typing import Optional, Union
 from pathlib import Path
 
 from .utils import CommandException, command
+from ..coretex import CustomDataset
 
 
 def cutadaptTrim(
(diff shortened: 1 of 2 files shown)


coretex/bioinformatics/__init__.py
```diff
@@ -52,19 +53,18 @@ def cutadaptTrim(
 
     args: list[str] = [
         "cutadapt",
-        "-o", forwardOutput,
+        "-o", str(forwardOutput),
```
dule1322, line 56: do it like this for all of them `if isinstance(forwardOutput, Path): forwardOutput = str(forwardOutput`. We don't wanna convert it to `str` if it is already `str` object

coretex/bioinformatics/__init__.py
```diff
@@ -62,9 +76,21 @@ def cutadaptTrim(
...
+    args.append(forwardFile)
     if reverseFile is not None:
         args.append(reverseFile)
 
-    args.append(forwardFile)
-
     command(args)
+
+
+def isPairedEnd(dataset: CustomDataset) -> bool:
```
dule1322, line 86: add documentation

PR #65, opened 2023-09-18
CTX-4396: Made it so that in case of "interval" x type of a metric, t…

VukManojlovic:
…he timestamp will equel x

### DIFF
--- a/coretex/coretex/experiment/experiment.py
+++ b/coretex/coretex/experiment/experiment.py
@@ -28,7 +28,7 @@ import json
 
 from .artifact import Artifact
 from .status import ExperimentStatus
-from .metrics import Metric
+from .metrics import Metric, MetricType
 from .parameter import validateParameters, parameter_factory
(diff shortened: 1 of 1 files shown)


coretex/coretex/experiment/experiment.py
```diff
@@ -306,12 +306,21 @@ def submitMetrics(self, metricValues: Dict[str, Tuple[float, float]]) -> bool:
...
-            "x": value[0],
-            "y": value[1]
-        } for key, value in metricValues.items()]
+        metrics : List[Dict[str, Any]] = []
+
+        timestamp = time.time()
+
+        for key, value in metricValues.items():
+            for metric in self.metrics:
+                if metric.name == key and metric.xType == MetricType.interval.name:
```
dule1322, line 315: Can double for-loop be avoided here? You can check if metric is in dict this way `metric.name in metricValues`

PR #68, opened 2023-09-20
CTX-4714: Increase requests connection pool size to match processor core count

dule1322:


igorperic17:
Interesting... Can you paste logs here to see the error this fixes? I'm curious. :D

dule1322:
> Interesting... Can you paste logs here to see the error this fixes? I'm curious. :D

It's a warning and it happens when more requests are asynchronously sent than what the connection pool size is so new connects are created but when requests finish they are not stored inside the pool since it might already be full so they get closed.

PR #71, opened 2023-09-25
CTX-4148: Parameters can now be overriden via command line arguments

VukManojlovic:


coretex/_task/_local.py
```diff
@@ -147,11 +167,15 @@ def processLocal(args: Optional[List[str]] = None) -> Tuple[int, TaskCallback]:
     if not os.path.exists("experiment.config"):
         raise FileNotFoundError(">> [Coretex] \"experiment.config\" file not found")
 
+    parameters = _readTaskRunConfig()
```
dule1322, line 170: Can extra parameters be passed in to `_readTaskRunConfig` function?

coretex/_task/_local.py
```diff
@@ -101,6 +112,15 @@ def configure(self) -> None:
         self.add_argument("--name", nargs = "?", type = str, default = None)
         self.add_argument("--description", nargs = "?", type = str, default = None)
 
+        for parameter in self.parameters:
+            if parameter.dataType in [ParameterType.dataset, ParameterType.enum, ParameterType.enumList, ParameterType.imuVectors]:
+                self.add_argument(f"--{parameter.name}", nargs = "?", type = parameter.overrideValue, default = None)
+            elif parameter.dataType.value.startswith("list"):
+                parameter = cast(BaseListParameter,  parameter)
```
dule1322, line 119: Instead of checking if `dataType` value starts with a list string and using `cast` a simple check can be performed via `isinstance(object, type)` and in a case where this is true python linter will automatically cast that object into the specified type and treat it as such.

coretex/_task/_local.py
```diff
@@ -86,13 +88,22 @@ def onCleanUp(self) -> None:
 
 class LocalArgumentParser(Tap):
 
+    parameters: List[BaseParameter]
```
dule1322, line 91: This can be removed

coretex/_task/_local.py
```diff
@@ -86,13 +88,22 @@ def onCleanUp(self) -> None:
...
     password: Optional[str]
 
     projectId: int
     name: Optional[str]
     description: Optional[str]
 
+    def __init__(self, parameters: List[BaseParameter]) -> None:
+        self.parameters = parameters
+        for parameter in parameters:
+            setattr(self, parameter.name, None)
```
dule1322, line 103: Why is `setattr` used here?

coretex/_task/_local.py
```diff
@@ -147,11 +176,19 @@ def processLocal(args: Optional[List[str]] = None) -> Tuple[int, TaskCallback]:
     if not os.path.exists("experiment.config"):
```
dule1322, line 176: This check should be moved alongside `_readTaskRunConfig` function

coretex/_task/_local.py
```diff
@@ -116,16 +136,25 @@ def _readTaskRunConfig() -> List[BaseParameter]:
...
 
-    parameterValidationResults = validateParameters(parameters, verbose = True)
-    if not all(parameterValidationResults.values()):
-        # Using this to make the parameter errors more readable without scrolling through the console
-        sys.exit(1)
-
     return parameters
 
 
+def _getParsedParameterValue(
```
dule1322, line 142: This function should be moved to the `LocalArgumentParser`

coretex/_task/_local.py
```diff
@@ -116,16 +136,25 @@ def _readTaskRunConfig() -> List[BaseParameter]:
...
+    parsedParameter = getattr(parser, parameterName)
+    if parsedParameter is None:
+        return default
+
+    return parsedParameter
+
+
 def processLocal(args: Optional[List[str]] = None) -> Tuple[int, TaskCallback]:
-    parser, unknown = LocalArgumentParser().parse_known_args(args)
+    parameters = _readTaskRunConfig()
```
dule1322, line 156: `_readTaskRunConfig` should also be moved to `LocalArgumentParser` as a `classmethod`

coretex/entities/task_run/parameter/parameters/list_enum_parameter.py
```diff
@@ -53,3 +55,12 @@ def parseValue(self, type_: ProjectType) -> Optional[Any]:
             return None
 
         return [options[value] for value in selected]
+
+    def overrideValue(self, value: Optional[Any]) -> Optional[Any]:
+        if value is None:
+            return None
+
+        try:
+            return json.loads(value.replace("'", "\""))
```
dule1322, line 64: Enum list parameter value should be a list of strings, why is it being treated as a json object?

dule1322 (changes requested):

coretex/entities/task_run/parameter/parameters/list_enum_parameter.py
VukManojlovic: The command line input will be gotten as a string ("['str1', 'str2', 'str3']"), which is most effectively cast into the corresponding json like object (list), using json.loads

dule1322: `tap` (Typed Argument Parser) supports lists as parameters, why do we have to manually parse them?

coretex/_task/_local.py
VukManojlovic: To dynamically create attributes for the LocalArgumentParser class, which are populated with parse_known_args, and then read with getattr in _getParsedParameterValue

coretex/entities/task_run/parameter/parameters/list_enum_parameter.py
VukManojlovic: Lists are parsed in a way that does not work for our context.
For example a list of strings will get parsed like this:
--parameter one two    ->    [["o", "n", "e"], ["t", "w", "o"]]

dule1322: This is not true, `--parameterName first second third` gets parsed into `["first", "second", "third"]`

VukManojlovic: That is not what I encountered. The desired results were only gotten by using nargs = "+" (meaning capture 1 and more passed args) with type = {non_list_type} (like int)

coretex/_task/_local.py
```diff
@@ -101,31 +110,51 @@ def configure(self) -> None:
...
 
-    parameterValidationResults = validateParameters(parameters, verbose = True)
-    if not all(parameterValidationResults.values()):
-        # Using this to make the parameter errors more readable without scrolling through the console
-        sys.exit(1)
+        return parameters
 
-    return parameters
+    @classmethod
+    def _getParsedParameterValue(
```
dule1322, line 141: This should not be a "classmethod" or a "staticmethod"

dule1322 (changes requested):

coretex/_task/_local.py
dule1322: Add a comment which describes why was this approach used

coretex/_task/_local.py
```diff
@@ -30,8 +30,10 @@
 from ._base_callback import TaskCallback
 from ._worker import TaskRunWorker
 from .. import folder_manager
-from ..entities import TaskRun, TaskRunStatus, BaseParameter, validateParameters, parameter_factory
+from ..entities import TaskRun, TaskRunStatus, BaseParameter, BaseListParameter, validateParameters, parameter_factory, ParameterType
 from ..networking import networkManager
+from ..entities.task_run.parameter.parameter_type import ParameterType
```
dule1322, line 35: Delete this import

coretex/_task/_local.py
```diff
@@ -30,8 +30,10 @@
...
+
```
dule1322, line 36: There are 3 newlines after the imports section

dule1322 (changes requested):

dule1322 (approved):

PR #75, opened 2023-09-28
CTX-4484: Samples will now be redownloaded if they have been modified since last download

VukManojlovic:
When a sample is downloaded, it's last modified time is updated to current time
Samples will be redownloaded if the last time modified of the downloading sample is newer then the last modified time of the one stored locally, if there is a locally cached version of the sample

### DIFF
--- a/coretex/entities/sample/network_sample.py
+++ b/coretex/entities/sample/network_sample.py
@@ -21,6 +21,8 @@ from datetime import datetime
 from pathlib import Path
 
 import os
+import time
+import logging
 
 from .sample import Sample
 from ..project import ProjectType
@@ -125,12 +127,15 @@ class NetworkSample(Generic[SampleDataType], Sample[SampleDataType], NetworkObje
         """
 
         if not self.zipPath.exists():
-            return False
+            raise FileNotFoundError(f">> [Coretex] Sample file could not be found at {self.zipPath}. Cannot check if file has been modified since last download")
 
-        sampleJson = self.fetchById(self.id).encode()
-        lastModified = sampleJson["storage_last_modified"]
+        #sampleJson = self.fetchById(self.id).encode()
+        #lastModified = sampleJson["storage_last_modified"]
+
+        lastModified = self.encode()["storage_last_modified"]
 
         if lastModified is None:
+            logging.error(f">> storage_last_modified is None for sample {self.name}")
             return False
 
         lastModifiedNew = datetime.strptime(lastModified, DATE_FORMAT)
@@ -148,7 +153,7 @@ class NetworkSample(Generic[SampleDataType], Sample[SampleDataType], NetworkObje
             bool -> False if response is failed, True otherwise
         """
 
-        if self.modifiedSinceLastDownload():
+        if self.zipPath.exists() and self.modifiedSinceLastDownload():
             ignoreCache = True
 
         response = networkManager.sampleDownload(
@@ -166,6 +171,9 @@ class NetworkSample(Generic[SampleDataType], Sample[SampleDataType], NetworkObje
 
                 sampleHardLinkPath.unlink()
                 os.link(self.zipPath, sampleHardLinkPath)
+
+        if Path(self.zipPath).exists():
+            os.utime(self.zipPath, (os.stat(self.zipPath).st_atime, time.time()))
 
         return not response.hasFailed()
 
--- a/coretex/entities/sample/network_sample.py
+++ b/coretex/entities/sample/network_sample.py
@@ -127,6 +127,10 @@ class NetworkSample(Generic[SampleDataType], Sample[SampleDataType], NetworkObje
             Returns
             -------
             bool -> False if sample has not changed since last download, True otherwise
+
+            Raises
+            ------
+            FileNotFoundError -> sample file cannot be found
         """
 
         if not self.zipPath.exists():
@@ -164,9 +168,7 @@ class NetworkSample(Generic[SampleDataType], Sample[SampleDataType], NetworkObje
 
                 sampleHardLinkPath.unlink()
                 os.link(self.zipPath, sampleHardLinkPath)
-
-        if self.zipPath.exists():
-            os.utime(self.zipPath, (os.stat(self.zipPath).st_atime, time.time()))
+                os.utime(self.zipPath, (os.stat(self.zipPath).st_atime, time.time()))
 
         return not response.hasFailed()
 
--- a/coretex/entities/sample/network_sample.py
+++ b/coretex/entities/sample/network_sample.py
@@ -17,6 +17,7 @@
 
 from typing import Any, TypeVar, Optional, Generic, Dict, Union
 from typing_extensions import Self
+from datetime import datetime
 from pathlib import Path
 
 import os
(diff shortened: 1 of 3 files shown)


coretex/entities/sample/network_sample.py
```diff
@@ -112,6 +114,31 @@ def _createSample(
...
+            Returns
+            -------
+            bool -> False if sample has not changed since last download, True otherwise
+        """
+
+        if not self.zipPath.exists():
+            return False
+
+        sampleJson = self.fetchById(self.id).encode()
+        lastModified = sampleJson["storage_last_modified"]
```
dule1322, line 131: Why is the sample being fetched and encoded here?

coretex/networking/network_manager_base.py
```diff
@@ -300,6 +302,9 @@ def sampleDownload(
             print(">> [Coretex] Retry count: {0}".format(retryCount))
             return self.sampleDownload(endpoint, destination, ignoreCache, parameters, retryCount + 1)
 
+        if Path(destination).exists():
+            os.utime(destination, (os.stat(destination).st_atime, time.time()))
```
dule1322, line 306: Does it make more sense to implement this in `Sample.download` function?

coretex/entities/sample/network_sample.py
```diff
@@ -112,6 +114,31 @@ def _createSample(
...
+            Checking if sample has been modified since last download, if the sample is already
+            stored locally
+
+            Returns
+            -------
+            bool -> False if sample has not changed since last download, True otherwise
+        """
+
+        if not self.zipPath.exists():
+            return False
```
dule1322, line 128: Should an exception be raised if the file does not exist?

coretex/entities/sample/network_sample.py
```diff
@@ -112,6 +114,31 @@ def _createSample(
...
+
+        sampleJson = self.fetchById(self.id).encode()
+        lastModified = sampleJson["storage_last_modified"]
+
+        if lastModified is None:
+            return False
```
dule1322, line 134: Why is it possible for `storage_last_modified` to be `None`?

dule1322 (changes requested):

coretex/entities/sample/network_sample.py
VukManojlovic: No. The file will only exist if was already downloaded before, so in case this is the first download it is expected for it not to exist

coretex/networking/network_manager_base.py
VukManojlovic: It may make more sense there

coretex/entities/sample/network_sample.py
VukManojlovic: This is a common value found in testing

coretex/entities/sample/network_sample.py
VukManojlovic: The needed field "storage_last_modified" gets sent only for fetch requests (as far as I could determine). It gets encoded afterward to read the "storage_last_modified" field from the original request json

dule1322: Sample should not be fetched again. Update implementation to not rely on fetching new samples

coretex/entities/sample/network_sample.py
dule1322: Let's update this function to raise an exception if the sample file does not exist

coretex/entities/sample/network_sample.py
dule1322: Has this been confirmed with the backend team?

VukManojlovic: Backend has confirmed that None is the expected value if the object has never been modified

coretex/entities/sample/network_sample.py
```diff
@@ -112,6 +116,34 @@ def _createSample(
...
+            bool -> False if sample has not changed since last download, True otherwise
+        """
+
+        if not self.zipPath.exists():
+            raise FileNotFoundError(f">> [Coretex] Sample file could not be found at {self.zipPath}. Cannot check if file has been modified since last download")
+
+        #sampleJson = self.fetchById(self.id).encode()
+        #lastModified = sampleJson["storage_last_modified"]
+
+        lastModified = self.encode()["storage_last_modified"]
```
dule1322, line 135: Add this to `NetworkSample` fields. Creating `KeyDescriptor` you can map this value to a `datetime` class automatically

coretex/entities/sample/network_sample.py
```diff
@@ -137,6 +172,9 @@ def download(self, ignoreCache: bool = False) -> bool:
                 sampleHardLinkPath.unlink()
                 os.link(self.zipPath, sampleHardLinkPath)
 
+        if Path(self.zipPath).exists():
```
dule1322, line 175: `self.zipPath` is already instance of `pathlib.Path`

coretex/entities/sample/network_sample.py
```diff
@@ -112,6 +116,34 @@ def _createSample(
...
+            Returns
+            -------
+            bool -> False if sample has not changed since last download, True otherwise
+        """
+
+        if not self.zipPath.exists():
+            raise FileNotFoundError(f">> [Coretex] Sample file could not be found at {self.zipPath}. Cannot check if file has been modified since last download")
+
+        #sampleJson = self.fetchById(self.id).encode()
+        #lastModified = sampleJson["storage_last_modified"]
```
dule1322, line 133: Remove comments

coretex/entities/sample/network_sample.py
dule1322: Let's remove this None check and modify this to not be nullable

dule1322 (changes requested):

coretex/entities/sample/network_sample.py
```diff
@@ -137,6 +165,9 @@ def download(self, ignoreCache: bool = False) -> bool:
                 sampleHardLinkPath.unlink()
                 os.link(self.zipPath, sampleHardLinkPath)
 
+        if self.zipPath.exists():
```
dule1322, line 168: This check is not needed, let's remove it and call `os.utime` as a part of the previous `if` clause which checks if response was successful or not.

coretex/entities/sample/network_sample.py
```diff
@@ -112,6 +119,24 @@ def _createSample(
...
+            Checking if sample has been modified since last download, if the sample is already
+            stored locally
+
+            Returns
+            -------
+            bool -> False if sample has not changed since last download, True otherwise
+        """
+
+        if not self.zipPath.exists():
+            raise FileNotFoundError(f">> [Coretex] Sample file could not be found at {self.zipPath}. Cannot check if file has been modified since last download")
```
dule1322, line 133: Documentation of functions should contain which exceptions might be raised by the function.

dule1322 (approved):

igorperic17:
Why is this in draft? Implementation seems OK and all discussions are resolved.

dule1322:
> Why is this in draft? Implementation seems OK and all discussions are resolved.

Backend changes were still in progress last week. They were finished on Friday but Vuk was off that day. As soon as Vuk tests his changes and they work ok we can merge this.

VukManojlovic:
> > Why is this in draft? Implementation seems OK and all discussions are resolved.
> 
> Backend changes were still in progress last week. They were finished on Friday but Vuk was off that day. As soon as Vuk tests his changes and they work ok we can merge this.

The changes have been tested with successful results

PR #80, opened 2023-10-04
CTX-2505: Implemented new parameter type "range"

VukManojlovic:


coretex/entities/task_run/parameter/parameters/__init__.py
```diff
@@ -23,9 +23,10 @@
...
 from .imu_vectors_parameter import IMUVectorsParameter
 from .enum_parameter import EnumParameter
+from .range_parameter import RangeParameter
 from .list_int_parameter import ListIntParameter
 from .list_float_parameter import ListFloatParameter
 from .list_str_parameter import ListStrParameter
 from .list_dataset_parameter import ListDatasetParameter
 from .list_model_parameter import ListModelParameter
-from .list_enum_parameter import ListEnumParameter
+from .list_enum_parameter import ListEnumParameter
```
dule1322, line 32: Why was newline removed here?

coretex/entities/task_run/parameter/parameters/range_parameter.py
```diff
@@ -0,0 +1,74 @@
...
+#     along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+from typing import Any, List, Dict, Optional, Tuple
+
+from ..base_parameter import BaseParameter
+from ..utils import validateRangeStructure
+from ....project import ProjectType
+
+
+class RangeParameter(BaseParameter[Dict[str, Any]]):
```
dule1322, line 25: Since we know the value of `range` parameter keys we can use them here instead of `Any`

coretex/entities/task_run/parameter/parameters/range_parameter.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def validate(self) -> Tuple[bool, Optional[str]]:
+        if not isinstance(self.value, dict):
+            return False, None
+
+        isValid, message = validateRangeStructure(self.name, self.value, self.required)
+        if not isValid:
+            return isValid, message
+
+        if self.value["to"] > 2^32:
```
dule1322, line 39: Let's extract `2^32` to a constant

coretex/entities/task_run/parameter/parameters/range_parameter.py
```diff
@@ -0,0 +1,74 @@
...
+            if len(args) not in [2, 3]:
+                return None
+
+            self.value["from"] = int(args[0])
+            self.value["to"] = int(args[1])
+            self.value["step"] = int(args[2]) if len(args) == 3 else 1
+
+            return self.value
+        except ValueError:
+            return None
```
dule1322, line 74: If an error is raised here will `TypedArgumentParser` detect it and display that error? This would help us output the error to user on what went wrong with the value which was entered

coretex/entities/task_run/parameter/utils.py
```diff
@@ -30,3 +30,18 @@ def validateEnumStructure(name: str, value: Optional[Any], required: bool) -> Tu
...
+
+def validateRangeStructure(name: str, value: Dict[str, Any], required: bool) -> Tuple[bool, Optional[str]]:
+    if len(value) != 3 or "from" not in value or "to" not in value or "step" not in value:
+        keys = ", ".join(value.keys())
+        return False, f"Range parameter \"{name}\" must contain only \"from\",  \"to\" and \"step\" properties, but it contains \"{keys}\""
+
+    if any(element is None for element in value.values()) and required:
+        return False, f"Elements of range parameter \"{name}\" must not be null"
+
+    if not all(isinstance(element, int) for element in value.values()) and required:
```
dule1322, line 43: `float` and `bool` are a subclasses of `int` in Python so `isintance(element, int)` will return `True` for cases where `element` is either a `float` or a `bool`. Can this be changed to check if `element` is `int` but not `int` subclasses?

dule1322 (changes requested):

coretex/entities/task_run/parameter/parameters/__init__.py
VukManojlovic: It was done by mistake

coretex/entities/task_run/parameter/parameters/range_parameter.py
VukManojlovic: I couldn't find a simple way to display the errors, without an exception being thrown

coretex/entities/task_run/parameter/utils.py
```diff
@@ -30,3 +30,21 @@ def validateEnumStructure(name: str, value: Optional[Any], required: bool) -> Tu
...
+def validateRangeStructure(name: str, value: Dict[str, Any], required: bool) -> Tuple[bool, Optional[str]]:
+    if len(value) != 3 or "from" not in value or "to" not in value or "step" not in value:
+        keys = ", ".join(value.keys())
+        return False, f"Range parameter \"{name}\" must contain only \"from\",  \"to\" and \"step\" properties, but it contains \"{keys}\""
+
+    if any(element is None for element in value.values()) and required:
+        return False, f"Elements of range parameter \"{name}\" must not be null"
+
+    try:
+        if not all(type(element)(int(element)) == element for element in value.values()) and required:
```
dule1322, line 44: To check if the type is just integer you can use `type(element) is int`

dule1322 (approved):

PR #81, opened 2023-10-04
CTX-4586: Made experiment.config optional

VukManojlovic:


coretex/_task/_local.py
```diff
@@ -34,6 +34,9 @@
 from ..networking import networkManager
 
 
+EXPERIMENT_CONGIF_PATH = "./experiment.config"
```
dule1322, line 37: Let's make this an instance of `pathlib.Path` class like so: `Path(".", "experiment.config")`

coretex/_task/_local.py
```diff
@@ -120,11 +123,12 @@ def configure(self) -> None:
...
-        if not os.path.exists("experiment.config"):
-            raise FileNotFoundError(">> [Coretex] \"experiment.config\" file not found")
-
         parameters: List[BaseParameter] = []
-        with open("./experiment.config", "rb") as configFile:
+
+        if not os.path.exists(EXPERIMENT_CONGIF_PATH):
+            return []
+
+        with open(EXPERIMENT_CONGIF_PATH, "rb") as configFile:
```
dule1322, line 131: After updating `EXPERIMENT_CONGIF_PATH` to be a `pathlib.Path` we can directly call `exists` and `open` methods on it.

dule1322, line 131: As a mode for opening the config file is it needed to use `"rb"` or is it possible to use just `"r"`?

dule1322 (changes requested):

dule1322 (approved):

PR #82, opened 2023-10-05
CTX-4588: Added dynamic entry point functionality

VukManojlovic:


coretex/entities/task_run/task_run.py
```diff
@@ -156,6 +157,7 @@ def _keyDescriptors(cls) -> Dict[str, KeyDescriptor]:
         descriptors["projectType"] = KeyDescriptor("project_task", ProjectType)
         descriptors["taskId"] = KeyDescriptor("sub_project_id")
         descriptors["taskName"] = KeyDescriptor("sub_project_name")
+        descriptors["entryPoint"] = KeyDescriptor("entry_point")
```
dule1322, line 160: There's no need for this. `KeyDescriptor` is only needed if a json field needs a custom type or a custom name. In this case `inflection` module will automatically convert `entry_point` into `entryPoint` so this can be removed.

dule1322 (changes requested):
Waiting for KeyDescriptor changes

dule1322 (approved):

PR #85, opened 2023-10-10
CTX-4148: Fixed boolean parameter not getting parsed properly

VukManojlovic:


coretex/entities/task_run/parameter/parameters/bool_parameter.py
```diff
@@ -25,3 +27,18 @@ class BoolParameter(BaseParameter[bool]):
...
+        if value is None:
+            return None
+
+        try:
+            if value.lower() == "true":
+                return True
+            elif value.lower() == "false":
+                return False
+            else:
+                raise ValueError("Could not recognise parsed value as boolean. Only accepted options are \"true\" and \"false\" (case insensitive)")
```
dule1322, line 41: There is no need to have this in `else`

PR #93, opened 2023-11-03
CTX-4915: Fixed centroid calculation

VukManojlovic:
- Changed np.asarray(image), in extractSegmentationMask, to np.array(image) as the former would create a non writable ndarray, which would throw an error as that array is written into in extractBinaryMask
- In the centroid calculation, the top left coordinate (minX and minY) is counted twice in the averaging as it appears twice in the segmentations (indices 0, 1 and 8, 9)

### DIFF
--- a/coretex/entities/annotation/image/coretex_format.py
+++ b/coretex/entities/annotation/image/coretex_format.py
@@ -121,7 +121,7 @@ class CoretexSegmentationInstance(Codable):
             draw = ImageDraw.Draw(image)
             draw.polygon(toPoly(segmentation), fill = 1)
 
-        return np.asarray(image)
+        return np.array(image)
 
     def extractBinaryMask(self, width: int, height: int) -> np.ndarray:
         """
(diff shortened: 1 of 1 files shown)


coretex/entities/annotation/image/coretex_format.py
```diff
@@ -156,10 +156,10 @@ def centroid(self) -> Tuple[int, int]:
 
         flattenedSegmentations = [element for sublist in self.segmentations for element in sublist]
 
-        listCX = [value for index, value in enumerate(flattenedSegmentations) if index % 2 == 0]
+        listCX = [value for index, value in enumerate(flattenedSegmentations) if index % 2 == 0 and index < 8]
```
dule1322, line 159: There can be N number of points in the `flattenedSegmentations` array so an `index` of `8` should not be used as a condition, the last point should be discarded from `flattenedSegmentations` array.

dule1322 (changes requested):

coretex/entities/annotation/image/coretex_format.py
```diff
@@ -154,7 +154,8 @@ def centroid(self) -> Tuple[int, int]:
             Tuple[int, int] -> x, y coordinates of centroid
         """
 
-        flattenedSegmentations = [element for sublist in self.segmentations for element in sublist]
+        flattenedSegmentations = [element for sublist in self.segmentations for i,  element in enumerate(sublist)\
+                                  if i < len(sublist) - 2]
```
dule1322, line 158: This was fixed in the annotation tool for coretex, so we no longer need to have the `if` part of the list comprehension

dule1322 (changes requested):

dule1322 (approved):

PR #100, opened 2023-12-01
CTX-2755: Fix git.Repo.ignored() hitting argument limit

VukManojlovic:


dule1322:
This does not fix the issue. Any kind of project which has large number of files will fail this.

coretex/entities/task_run/utils.py
```diff
@@ -90,20 +90,28 @@ def getDefaultEntryPoint() -> Optional[str]:
     return None
 
 
+def chunks(lst: list, n: int) -> list:
+    for i in range(0, len(lst), n):
+        yield lst[i:i + n]
```
dule1322, line 95: What happens with remainder if `len(lst) % n != 0`?

dule1322 (changes requested):

coretex/entities/task_run/utils.py
VukManojlovic: Screenshot 2023-12-08 at 16 51 47
The remainder gets returned as a list of length len(lst) % n

dule1322 (approved):

PR #101, opened 2023-12-04
CTX-5033: Allow integers instead of MetricType enum as type arguments for Metric.create

VukManojlovic:


coretex/entities/task_run/metrics/metric.py
```diff
@@ -68,6 +68,12 @@ def create(
                 range in which values will be displayed for y axis
         """
 
+        if isinstance(xType, int):
+            xType = createMetricType(xType)
```
dule1322, line 72: There is no need for a specific function for creating enum from type. You can just do `xType = MetricType(xType)`

dule1322 (changes requested):

coretex/entities/task_run/metrics/metric.py
```diff
@@ -44,9 +44,9 @@ def create(
         cls,
         name: str,
         xLabel: str,
-        xType: MetricType,
+        xType: Union[MetricType, int],
         yLabel: str,
-        yType: MetricType,
+        yType: Union[MetricType, int],
```
dule1322, line 49: Update the function documentation accordingly.

dule1322 (changes requested):

PR #103, opened 2023-12-05
CTX-4911: Automatic artifact upload

dule1322:


dule1322:
Draft until CTX-2755 gitignore functionality is properly fixed. This is needed for ignoring files which are inside .gitignore when uploading artifacts

PR #114, opened 2024-01-15
CTX-5263: Split log files into multiple folders

VukManojlovic:


coretex/_task/base_callback.py
```diff
@@ -50,7 +50,10 @@ def onStart(self) -> None:
         # Call "kill -30 task_run_process_id" to dump current stack trace of the TaskRun into the file
         # 30 == signal.SIGUSR1
         # Only works on *nix systems
-        faultHandlerPath = folder_manager.logs / f"task_run_stacktrace_{self._taskRun.id}_{datetime.now().strftime(DATE_FORMAT)}.log"
+        taskRunLogs = folder_manager.logs / "runs" / str(self._taskRun.id)
+        taskRunLogs.mkdir(parents = True, exist_ok = True)
```
dule1322, line 54: Let's extract `folder_manager.logs / "runs"` to `folder_manager` module so we can access it like this `folder_manager.runsLogDirectory`

coretex/folder_manager.py
```diff
@@ -68,6 +68,8 @@ def _createFolder(name: str) -> Path:
 temp             = _createFolder("temp")
 _artifactsFolder = _createFolder("artifacts")
 
+runsLogDirectory = logs / "runs"
```
dule1322, line 71: call `mkdir` function here to make sure this dir will always exist

dule1322 (changes requested):

coretex/_task/worker/utils.py
```diff
@@ -43,7 +43,10 @@ def initializeLogger(taskRunId: int) -> None:
         style = "%",
     )
 
-    workerLogPath = folder_manager.logs / f"task_run_worker_{taskRunId}.log"
+    taskRunLogs = folder_manager.runsLogDirectory / str(taskRunId)
+    taskRunLogs.mkdir(parents = True, exist_ok = True)
```
dule1322, line 47: Let's create a function in `folder_manager` called `getRunLogsDir` which will have `taskRunId` as a parameter and return a directory for storing logs of that Task Run.

coretex/_logger.py
```diff
@@ -24,5 +24,8 @@
 
 
 def _initializeDefaultLogger() -> None:
-    logPath = folder_manager.logs / f"coretexpylib_{datetime.now().strftime(DATE_FORMAT)}.log"
+    ctxpylibLogs = folder_manager.logs / "coretexpylib"
+    ctxpylibLogs.mkdir(exist_ok = True)
```
dule1322, line 28: Let's also move this to `folder_manager` module

dule1322 (changes requested):

dule1322 (approved):

PR #128, opened 2024-02-12
CTX-5424: Fix for crash (accessing modelId might fail since modelId…

bogdant36:
… doesn't have to exist).


dule1322:
model id is not passed in as env variable when Node autoupdate happens

PR #129, opened 2024-02-12
CTX-4880: If gpu is specified in config, container will be restarted with gpu capability.

bogdant36:
Resolves CTX-4880


coretex/cli/resources/update_script_template.sh
```diff
@@ -89,7 +90,7 @@ function stop_node {{
 }}
 
 # Define function to start the node with the latest image
-function start_node {{
+start_node() {
```
dule1322, line 93: why was this changed?

PR #136, opened 2024-02-16
 CTX-5258: Added request timeouts to NetworkManager module

VukManojlovic:


coretex/networking/network_manager_base.py
```diff
@@ -407,7 +410,9 @@ def formData(
             headers = self._headers("multipart/form-data")
             del headers["Content-Type"]
 
-            return self.request(endpoint, RequestType.post, headers, body = params, files = filesData)
+            timeout = 3600 if endpoint in ["upload/chunk", "artifact/upload-file"] else REQUEST_TIMEOUT
```
dule1322, line 413: Do not check for endpoints directly since if another one which is uploading a file is used we need to manually enter it here. Use `files` parameter to decide if timeout should be increased or not. Also only read timeout should be increased in upload case right? Not connection timeout.

dule1322 (changes requested):

coretex/networking/network_manager_base.py
```diff
@@ -407,7 +410,9 @@ def formData(
             headers = self._headers("multipart/form-data")
             del headers["Content-Type"]
 
-            return self.request(endpoint, RequestType.post, headers, body = params, files = filesData)
+            timeout: Union[int, Tuple[int, int]] = (600, 300) if len(files) > 0 else REQUEST_TIMEOUT
```
dule1322, line 413: Can we extract the `timeout` type into a type alias since it is used in mutliple places?

dule1322 (changes requested):

coretex/networking/network_manager_base.py
```diff
@@ -407,7 +412,9 @@ def formData(
             headers = self._headers("multipart/form-data")
             del headers["Content-Type"]
 
-            return self.request(endpoint, RequestType.post, headers, body = params, files = filesData)
+            timeout: TimeoutType = (600, 300) if len(files) > 0 else REQUEST_TIMEOUT
```
dule1322, line 415: Why is the connection timeout for HTTP so long?

dule1322 (changes requested):

coretex/networking/network_manager_base.py
VukManojlovic: Testing on 128MB file uploads, 60s was just above the thresholds where the request would succeed. I 10x-ed this value to account for slower user upload speed

dule1322: Can we do another approach for this, as this is not optimal?

coretex/networking/network_manager_base.py
```diff
@@ -407,7 +412,16 @@ def formData(
             headers = self._headers("multipart/form-data")
             del headers["Content-Type"]
 
-            return self.request(endpoint, RequestType.post, headers, body = params, files = filesData)
+            timeout: TimeoutType
+            if len(files) == 0:
+                timeout = REQUEST_TIMEOUT
+            else:
+                if self.request(endpoint, RequestType.options, timeout = REQUEST_TIMEOUT).hasFailed():
+                    raise ConnectionError(">> [Coretex] Could not establish a connection with the server")
```
dule1322, line 420: If the options request failed raise `NetworkRequestError`.

coretex/networking/network_manager_base.py
```diff
@@ -407,7 +412,16 @@ def formData(
...
+
+                timeout = (600, 300)
```
dule1322, line 422: Timeout should not be set in case of upload if options request has passed

dule1322 (changes requested):

coretex/networking/network_manager_base.py
```diff
@@ -407,7 +412,17 @@ def formData(
...
             del headers["Content-Type"]
 
-            return self.request(endpoint, RequestType.post, headers, body = params, files = filesData)
+            timeout: TimeoutType
+            if len(files) == 0:
+                timeout = REQUEST_TIMEOUT
+            else:
+                response = self.request(endpoint, RequestType.options, timeout = REQUEST_TIMEOUT)
+                if response.hasFailed():
+                    raise NetworkRequestError(response, ">> [Coretex] Could not establish a connection with the server")
```
dule1322, line 421: There is no need to specify `>> [Coretex]` prefix here.

dule1322 (approved):

PR #147, opened 2024-03-05
CTX-5600: Implemented functionality to handle if container exists but is not running.

bogdant36:
Resolves CTX-5600

### DIFF
--- a/coretex/cli/modules/node.py
+++ b/coretex/cli/modules/node.py
@@ -65,8 +65,10 @@ def pull(image: str) -> None:
 
 
 def isRunning() -> bool:
-    return docker.containerExists(DOCKER_CONTAINER_NAME)
+    return docker.containerRunning(DOCKER_CONTAINER_NAME)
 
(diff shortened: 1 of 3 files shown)


coretex/cli/modules/node.py
```diff
@@ -132,15 +134,17 @@ def start(dockerImage: str, config: Dict[str, Any]) -> None:
         raise NodeException("Failed to start Coretex Node.")
 
 
-def stop() -> None:
+def stop(echoOutput: bool = True) -> None:
```
dule1322, line 137: Should there be a new function which removes network and container? It isn't very clear why is stop being used if a container is not running but it exists.

dule1322 (changes requested):

coretex/cli/modules/node.py
bogdant36: Now when I think a bit about it you are right it is confusing.

dule1322 (approved):

PR #157, opened 2024-03-12
Ctx 5526: Coretex run command initial implemetation.

bogdant36:
Ctx 5526: Coretex run command initial implemetation.


dule1322:
Please take a look at CTX-5526 task comments and update the command flow as specified.

PR #158, opened 2024-03-14
CTX-5467: Handling resource limitations inside docker desktop appli…

bogdant36:
…cation and notifying user about limitations.


coretex/utils/docker.py
```diff
@@ -123,3 +124,17 @@ def imageInspect(image: str) -> Dict[str, Any]:
         raise TypeError(f"Invalid function result type \"{type(jsonOutput[0])}\". Expected: \"dict\"")
 
     return jsonOutput[0]
+
+
+def dockerInfo() -> Tuple[int, int, int]:
+    _, output, _ = command(["docker", "info"], ignoreStdout = True, ignoreStderr = True)
```
dule1322, line 130: `docker info` command can be formatted to json, please do so instead of using regex to manually parse values

coretex/cli/modules/node.py
```diff
@@ -90,14 +99,17 @@ def start(dockerImage: str, config: Dict[str, Any]) -> None:
         if initScript is not None:
             volumes.append((str(initScript), "/script/init.sh"))
 
+        limitedCpus, limitedRam, limitedSwap = docker.dockerInfo()
+        validateResources(config, limitedCpus, limitedRam, limitedSwap)
```
dule1322, line 103: This can be expanded to a `validateNodeConfig` to validate if node is properly configured, not just for resources, but for everything

coretex/cli/modules/node.py
```diff
@@ -90,14 +99,17 @@ def start(dockerImage: str, config: Dict[str, Any]) -> None:
...
+
         docker.start(
             config_defaults.DOCKER_CONTAINER_NAME,
             dockerImage,
             config["allowGpu"],
-            config["nodeRam"],
-            config["nodeSwap"],
+            limitedRam if config["nodeRam"] >= limitedRam and limitedRam > 0 else config["limitedRam"],
```
dule1322, line 109: User should not be able to start the node at all if the configuration is not valid

coretex/cli/modules/node.py
dule1322: Validate the values entered for the resources during node config too, not just during node start

dule1322 (changes requested):

coretex/cli/modules/node.py
```diff
@@ -302,13 +366,17 @@ def configureNode(config: Dict[str, Any], verbose: bool) -> None:
...
+
 
 def initializeNodeConfiguration() -> None:
     config = loadConfig()
 
     if isNodeConfigured(config):
+        validateConfiguration(config)
         return
 
+
```
dule1322, line 379: Why was this newline added?

dule1322 (changes requested):

coretex/cli/modules/node.py
```diff
@@ -287,10 +307,24 @@ def configureNode(config: Dict[str, Any], verbose: bool) -> None:
...
+        cpuCount = clickPrompt("Enter the number of CPUs the container will use (press enter to use default)", config_defaults.DEFAULT_CPU_COUNT, type = int)
+        if cpuCount > cpuLimit:
+            stdEcho(f"WARNING: CPU limit in Docker Desktop ({cpuLimit}) is lower than the configured value ({config['cpuCount']}). Please adjust resource limitations in Docker Desktop settings.")
+            cpuCount = cpuLimit
+        config["cpuCount"] = cpuCount
+
+        if config_defaults.DEFAULT_RAM_MEMORY < config_defaults.MINIMUM_RAM_MEMORY:
+            errorEcho("Minimum Node RAM requirement is 6GB.")
+            return
```
dule1322, line 319: This should raise an exception

coretex/cli/modules/node.py
```diff
@@ -287,10 +307,24 @@ def configureNode(config: Dict[str, Any], verbose: bool) -> None:
...
+        nodeRam = clickPrompt("Node RAM memory limit in GB (press enter to use default)", config_defaults.DEFAULT_RAM_MEMORY, type = int)
+        if nodeRam > ramLimit:
+            stdEcho(f"WARNING: CPU limit in Docker Desktop ({cpuLimit}) is lower than the configured value ({config['cpuCount']}). Please adjust resource limitations in Docker Desktop settings.")
+            nodeRam = ramLimit
```
dule1322, line 323: It should not be allowed to configure Node with lower value compared to limit, this should either prompt user to enter the value again until a valid value is entered.

In RAM case where there is a hard minimum limit it should flat out refuse to configure/start the Node if there is not enough resources.

coretex/cli/modules/node.py
```diff
@@ -255,8 +255,28 @@ def _configureInitScript() -> str:
...
+        raise TypeError(f"Invalid config \"cpuCount\" field type \"{type(config['cpuCount'])}\". Expected: \"int\"")
+
+    if cpuLimit < config["cpuCount"]:
+        raise RuntimeError(f"Configuration not valid. CPU limit in Docker Desktop ({cpuLimit}) is lower than the configured value ({config['cpuCount']})")
+
+    if ramLimit < config["nodeRam"]:
+        raise RuntimeError(f"Configuration not valid. RAM limit in Docker Desktop ({ramLimit}) is lower than the configured value ({config['nodeRam']})")
+
+    if config_defaults.MINIMUM_RAM_MEMORY > config["nodeRam"]:
+        raise RuntimeError(f"Configuration not valid. Minimum Node RAM requirement ({config_defaults.MINIMUM_RAM_MEMORY}) is higher than the configured value ({config['nodeRam']})")
```
dule1322, line 274: Do not raise exceptions for these messages, print out all of them at once and return True/False if config is valid or not

coretex/cli/modules/node.py
```diff
@@ -236,6 +236,23 @@ def selectNodeMode(storagePath: str) -> Tuple[int, Optional[int]]:
     return availableNodeModes[selectedMode], None
 
 
+def promptRam(config: Dict[str, Any], ramLimit: int, retryCount: int = 0) -> int:
```
dule1322, line 239: `retryCount` is unused except in recursive function calls, it is safe to remove completely.

coretex/cli/modules/node.py
```diff
@@ -236,6 +236,23 @@ def selectNodeMode(storagePath: str) -> Tuple[int, Optional[int]]:
...
+    nodeRam: int = clickPrompt("Node RAM memory limit in GB (press enter to use default)", config_defaults.DEFAULT_RAM_MEMORY, type = int)
+
+    if ramLimit < config_defaults.MINIMUM_RAM_MEMORY:
+        raise RuntimeError(f"Minimum Node RAM requirement ({ramLimit}GB) is higher than your current Docker desktop RAM limit ({ramLimit}GB). Please adjust resource limitations in Docker Desktop settings to match Node requirements.")
+
+    if nodeRam > ramLimit:
+        errorEcho(f"WARNING: RAM limit in Docker Desktop ({ramLimit}GB) is lower than the configured value ({config['ramLimit']}GB). Please adjust resource limitations in Docker Desktop settings.")
```
dule1322, line 246: This is an error not a warning. When a warning happens it means that the situation which is a result of it does not affect the normal flow of the program, whilst when an error happens it means that the normal flow gets interrupted.

coretex/cli/modules/node.py
```diff
@@ -236,6 +236,23 @@ def selectNodeMode(storagePath: str) -> Tuple[int, Optional[int]]:
...
+        return promptRam(config, ramLimit, retryCount + 1)
+
+    if nodeRam < config_defaults.MINIMUM_RAM_MEMORY:
+        errorEcho(f"WARNING: Configured RAM ({config['ramLimit']}GB) is lower than the minimum Node RAM requirement ({config_defaults.MINIMUM_RAM_MEMORY}GB).")
```
dule1322, line 250: Here as well.

coretex/cli/modules/node.py
```diff
@@ -255,8 +272,47 @@ def _configureInitScript() -> str:
...
+
+    if ramLimit < config["nodeRam"]:
+        errorEcho(f"Configuration not valid. RAM limit in Docker Desktop ({ramLimit}GB) is lower than the configured value ({config['nodeRam']}GB)")
+        isValid = False
+
+    if ramLimit < config_defaults.MINIMUM_RAM_MEMORY:
+        errorEcho(f"Minimum Node RAM requirement ({config_defaults.MINIMUM_RAM_MEMORY}GB) is higher than your current Docker desktop RAM limit ({ramLimit}GB). Please adjust resource limitations in Docker Desktop settings to match Node requirements.")
+        isValid = False
+
+    if config_defaults.MINIMUM_RAM_MEMORY > config["nodeRam"]:
```
dule1322, line 306: Place `config` value on the left side of the `>` operator. We are comparing that value against something, not the other way around. Do this for all `if` statements in the function.

coretex/cli/modules/node.py
```diff
@@ -255,8 +272,47 @@ def _configureInitScript() -> str:
...
+        errorEcho(f"Configuration not valid. Minimum Node RAM requirement ({config_defaults.MINIMUM_RAM_MEMORY}GB) is higher than the configured value ({config['nodeRam']}GB)")
+        isValid = False
+
+    return isValid
+
+
 def configureNode(config: Dict[str, Any], verbose: bool) -> None:
     highlightEcho("[Node Configuration]")
+    cpuLimit, ramLimit = docker.getResourceLimits()
```
dule1322, line 315: Add newline above and below this line

coretex/cli/modules/node.py
```diff
@@ -276,7 +332,7 @@ def configureNode(config: Dict[str, Any], verbose: bool) -> None:
         config["image"] += f":latest-{tag}"
 
     config["storagePath"] = config_defaults.DEFAULT_STORAGE_PATH
-    config["nodeRam"] = config_defaults.DEFAULT_RAM_MEMORY
+    config["nodeRam"] = ramLimit if config_defaults.DEFAULT_RAM_MEMORY > ramLimit else config_defaults.DEFAULT_RAM_MEMORY
```
dule1322, line 335: Use `min` function here instead of ternary expression

coretex/cli/modules/node.py
```diff
@@ -307,7 +371,10 @@ def initializeNodeConfiguration() -> None:
     config = loadConfig()
 
     if isNodeConfigured(config):
-        return
+        if isConfigurationValid(config):
```
dule1322, line 374: Aren't these two (`isNodeConfigured` and `isConfigurationValid`) the same function?

coretex/utils/docker.py
```diff
@@ -123,3 +123,10 @@ def imageInspect(image: str) -> Dict[str, Any]:
         raise TypeError(f"Invalid function result type \"{type(jsonOutput[0])}\". Expected: \"dict\"")
 
     return jsonOutput[0]
+
+
+def getResourceLimits() -> Tuple[int, int]:
+    _, output, _ = command(["docker", "info", "--format", "{{json .}}"], ignoreStdout = True, ignoreStderr = True)
+    jsonOutput = json.loads(output)
+
+    return jsonOutput["NCPU"], round(jsonOutput["MemTotal"] / (1024**3), 0)
```
dule1322, line 132: Add spaces aroud pow operator

coretex/cli/modules/node.py
```diff
@@ -287,10 +343,18 @@ def configureNode(config: Dict[str, Any], verbose: bool) -> None:
...
-        config["nodeRam"] = clickPrompt("Node RAM memory limit in GB (press enter to use default)", config_defaults.DEFAULT_RAM_MEMORY, type = int)
+
+        cpuCount = clickPrompt("Enter the number of CPUs the container will use (press enter to use default)", config_defaults.DEFAULT_CPU_COUNT, type = int)
+        if cpuCount > cpuLimit:
+            stdEcho(f"WARNING: CPU limit in Docker Desktop ({cpuLimit}) is lower than the configured value ({config['cpuCount']}). Please adjust resource limitations in Docker Desktop settings.")
+            cpuCount = cpuLimit
+        config["cpuCount"] = cpuCount
+
+        nodeRam = promptRam(config, ramLimit)
+        config["nodeRam"] = nodeRam
```
dule1322, line 354: Why not directly assign `promptRam` return value to `config["nodeRam"]`?

coretex/cli/modules/node.py
```diff
@@ -287,10 +343,18 @@ def configureNode(config: Dict[str, Any], verbose: bool) -> None:
 
     if verbose:
         config["storagePath"] = clickPrompt("Storage path (press enter to use default)", config_defaults.DEFAULT_STORAGE_PATH, type = str)
-        config["nodeRam"] = clickPrompt("Node RAM memory limit in GB (press enter to use default)", config_defaults.DEFAULT_RAM_MEMORY, type = int)
+
+        cpuCount = clickPrompt("Enter the number of CPUs the container will use (press enter to use default)", config_defaults.DEFAULT_CPU_COUNT, type = int)
+        if cpuCount > cpuLimit:
+            stdEcho(f"WARNING: CPU limit in Docker Desktop ({cpuLimit}) is lower than the configured value ({config['cpuCount']}). Please adjust resource limitations in Docker Desktop settings.")
+            cpuCount = cpuLimit
+        config["cpuCount"] = cpuCount
```
dule1322, line 351: This logic is invalid, it should not default to `cpuLimit` in verbose mode. The user must never be able to enter incorrect value using CLI.

coretex/cli/commands/node.py
```diff
@@ -155,6 +155,7 @@ def config(verbose: bool) -> None:
 
 
 @click.group()
+@onBeforeCommandExecute(node_module.checkResourceLimitations)
```
dule1322, line 158: If the `checkResourceLimitations` command gets executed on every command from the `node` group doesn't that mean it will get executed when I call `coretex node config`, and if by any chance I have invalid configuration it will fail?

dule1322: This should call the same function for validating node configuration, and there should be only one function which does that to keep it straightforward, bug-prone, and easy to modify.

dule1322 (changes requested):

coretex/cli/modules/node.py
bogdant36: Do we want to keep prompting for ram forever or to end as soon as user enters incorrect value?

coretex/cli/commands/node.py
bogdant36: My idea was to prevent users that dont have enough resources that are mandatory (for example RAM) from using the node functionality since they will not be able to do anything.

coretex/cli/modules/node.py
bogdant36: Well right now, these 2 functions are different. isNodeConfigured() function is only checking for required fields for node in config and isConfigurationValid function is checking if each field has valid type and value (if must be checked). Do you want this to be changed and in what way?

coretex/cli/modules/utils.py
```diff
@@ -15,7 +15,7 @@
 #     You should have received a copy of the GNU Affero General Public License
 #     along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-from typing import List, Any, Optional, Callable
+from typing import List, Any, Dict, Optional, Callable
```
dule1322, line 18: This import is unused

coretex/cli/modules/node.py
```diff
@@ -236,6 +236,32 @@ def selectNodeMode(storagePath: str) -> Tuple[int, Optional[int]]:
     return availableNodeModes[selectedMode], None
 
 
+def promptCpu(config: Dict[str, Any], cpuLimit: int) -> int:
+    cpuCount: int = clickPrompt("Enter the number of CPUs the container will use (press enter to use default)", config_defaults.DEFAULT_CPU_COUNT, type = int)
```
dule1322, line 240: Default value should be minimum of `config_defaults.DEFAULT_CPU_COUNT` and `cpuLimit`.

coretex/cli/modules/node.py
```diff
@@ -236,6 +236,32 @@ def selectNodeMode(storagePath: str) -> Tuple[int, Optional[int]]:
...
+
+    if cpuCount > cpuLimit:
+        errorEcho(f"ERRROR: CPU limit in Docker Desktop ({cpuLimit}) is lower than the specified value ({cpuCount})")
+        return promptCpu(config, cpuLimit)
+
+    return cpuCount
+
+def promptRam(config: Dict[str, Any], ramLimit: int) -> int:
+    nodeRam: int = clickPrompt("Node RAM memory limit in GB (press enter to use default)", config_defaults.DEFAULT_RAM_MEMORY, type = int)
```
dule1322, line 249: Default value should be minimum of `config_defaults.DEFAULT_RAM_MEMORY` and `ramLimit`.

coretex/cli/modules/node.py
dule1322: We can combine the 2 functions into a single one

coretex/cli/modules/node.py
dule1322: Forever

coretex/cli/modules/node.py
```diff
@@ -236,6 +236,32 @@ def selectNodeMode(storagePath: str) -> Tuple[int, Optional[int]]:
...
+
+    if ramLimit < config_defaults.MINIMUM_RAM_MEMORY:
```
dule1322, line 251: Shouldn't this be checked way earlier if the user does not have enough RAM?

coretex/cli/modules/node.py
bogdant36: After short discussion we decided to keep two functions for now.

coretex/cli/modules/node.py
```diff
@@ -236,6 +236,29 @@ def selectNodeMode(storagePath: str) -> Tuple[int, Optional[int]]:
     return availableNodeModes[selectedMode], None
 
 
+def promptCpu(config: Dict[str, Any], cpuLimit: int) -> int:
+    cpuCount: int = clickPrompt("Enter the number of CPUs the container will use (press enter to use default)", cpuLimit, type = int)
+
+    if cpuCount > cpuLimit:
+        errorEcho(f"ERRROR: CPU limit in Docker Desktop ({cpuLimit}) is lower than the specified value ({cpuCount})")
```
dule1322, line 243: `ERRROR` is misspelled

coretex/cli/modules/node.py
```diff
@@ -236,6 +236,29 @@ def selectNodeMode(storagePath: str) -> Tuple[int, Optional[int]]:
...
+        return promptCpu(config, cpuLimit)
+
+    return cpuCount
+
```
dule1322, line 247: Top-level definitions should be separated by 2 newlines

coretex/cli/modules/node.py
```diff
@@ -236,6 +236,29 @@ def selectNodeMode(storagePath: str) -> Tuple[int, Optional[int]]:
     return availableNodeModes[selectedMode], None
 
 
+def promptCpu(config: Dict[str, Any], cpuLimit: int) -> int:
+    cpuCount: int = clickPrompt("Enter the number of CPUs the container will use (press enter to use default)", cpuLimit, type = int)
```
dule1322, line 240: Let's display the max value for CPU as a part of the prompt.

coretex/cli/modules/node.py
```diff
@@ -236,6 +236,29 @@ def selectNodeMode(storagePath: str) -> Tuple[int, Optional[int]]:
...
+
+    if cpuCount > cpuLimit:
+        errorEcho(f"ERRROR: CPU limit in Docker Desktop ({cpuLimit}) is lower than the specified value ({cpuCount})")
+        return promptCpu(config, cpuLimit)
+
+    return cpuCount
+
+def promptRam(config: Dict[str, Any], ramLimit: int) -> int:
+    nodeRam: int = clickPrompt("Node RAM memory limit in GB (press enter to use default)", ramLimit, type = int)
```
dule1322, line 249: Let's display the min and max values as a part of the prompt. Min value is 6 GB, whilst the max is either ramLimit (if it is set), or the total amount of RAM available for the system.

coretex/cli/modules/node.py
```diff
@@ -307,7 +375,10 @@ def initializeNodeConfiguration() -> None:
     config = loadConfig()
 
     if isNodeConfigured(config):
-        return
+        if isConfigurationValid(config):
```
dule1322, line 378: `if firstCondition and secondCondition and thirdCondition` is equivalent to `if firstCondition: if secondCondition: if thirdCondition` functionaly, but the former is much easier to read as it does not have unnecessary nesting. Refactor this whole if-else branching as there is no need for the else block to exist at all.

dule1322 (changes requested):

coretex/cli/modules/node.py
```diff
@@ -236,6 +236,30 @@ def selectNodeMode(storagePath: str) -> Tuple[int, Optional[int]]:
...
+
+    if cpuCount > cpuLimit:
+        errorEcho(f"ERROR: CPU limit in Docker Desktop ({cpuLimit}) is lower than the specified value ({cpuCount})")
+        return promptCpu(config, cpuLimit)
+
+    return cpuCount
+
+
+def promptRam(config: Dict[str, Any], ramLimit: int) -> int:
+    nodeRam: int = clickPrompt(f"Node RAM memory limit in GB (Minimum: 6GB, Maximum: {ramLimit})(press enter to use default)", ramLimit, type = int)
```
dule1322, line 250: space missing between brackets

coretex/cli/modules/node.py
```diff
@@ -236,6 +236,30 @@ def selectNodeMode(storagePath: str) -> Tuple[int, Optional[int]]:
...
+
+    if cpuCount > cpuLimit:
+        errorEcho(f"ERROR: CPU limit in Docker Desktop ({cpuLimit}) is lower than the specified value ({cpuCount})")
+        return promptCpu(config, cpuLimit)
+
+    return cpuCount
+
+
+def promptRam(config: Dict[str, Any], ramLimit: int) -> int:
+    nodeRam: int = clickPrompt(f"Node RAM memory limit in GB (Minimum: 6GB, Maximum: {ramLimit}) (press enter to use default)", ramLimit, type = int)
```
dule1322, line 250: Why is minimum value not extracted from a constant? Also minimum has GB notation while maximum will just display a number without any notation

dule1322 (changes requested):

dule1322 (approved):

PR #160, opened 2024-03-19
CTX-5384: Added Utils method for formatting bytes to human readable format.

jovicazaric:


coretex/utils/number.py
```diff
@@ -39,3 +39,25 @@ def mathematicalRound(value: float, decimalPlaces: int) -> float:
...
+def formatBytes(value: int, precision = 2):
+    """
+    Format a size in bytes into a human-readable format.
+
+    Parameters:
+        value (int): The size in bytes.
+        precision (int): Number of decimal places (default is 2).
+
+    Returns:
+        str: Formatted string with size and appropriate unit.
```
dule1322, line 53: Use the same formatting as we do for documentation of other functions

coretex/utils/number.py
```diff
@@ -39,3 +39,25 @@ def mathematicalRound(value: float, decimalPlaces: int) -> float:
     places = Decimal(10) ** -decimalPlaces
 
     return float(decimal.quantize(places, rounding = ROUND_HALF_UP))
+
+
+def formatBytes(value: int, precision = 2):
```
dule1322, line 44: Explicitly annotate type of `precision` and return type of the function.

coretex/utils/number.py
```diff
@@ -39,3 +39,25 @@ def mathematicalRound(value: float, decimalPlaces: int) -> float:
...
+
+    Parameters:
+        value (int): The size in bytes.
+        precision (int): Number of decimal places (default is 2).
+
+    Returns:
+        str: Formatted string with size and appropriate unit.
+    """
+
+    suffixes = ['B', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB']
```
dule1322, line 56: Use double quotes for strings.

coretex/utils/number.py
jovicazaric: Updated

coretex/utils/number.py
jovicazaric: Added.

coretex/utils/number.py
jovicazaric: Updated

PR #161, opened 2024-03-26
CTX-5143: CLI - coretex project create & edit commands implementation.

bogdant36:
Resolves CTX-5143.


coretex/cli/commands/project.py
```diff
@@ -20,47 +20,106 @@
 import click
 
 from ..modules import ui, project_utils, utils, user
-from ...entities import Project
-from ...networking import NetworkRequestError
-from ...configuration import loadConfig, saveConfig
+from ..modules.ui import clickPrompt, errorEcho, successEcho, progressEcho
```
dule1322, line 23: Let's use `from modules import ui` and then call these functions with `ui` prefix.

coretex/cli/commands/project.py
```diff
@@ -20,47 +20,106 @@
...
+from ...entities import Project, ProjectType, ProjectVisibility
+from ...networking import NetworkRequestError, EntityNotCreated, networkManager, RequestFailedError
+from ...configuration import loadConfig
 
 
-def selectProject(projectId: int) -> None:
+@click.command()
+@click.option("--project", "-p", type = str, help = "Project name")
```
dule1322, line 30: Call the option `--name` instead of `--project` when creating a project

coretex/cli/commands/project.py
```diff
@@ -20,47 +20,106 @@
...
+        project = Project.createProject(name, ProjectType(type), description)
+
+        if project is not None:
+            ui.successEcho(f"Project \"{name}\" created successfully.")
+            selectNewProject = clickPrompt("Do you want to select new project as default? (Y/n)", type = bool, default = True)
+
+            if selectNewProject:
+                project_utils.selectProject(project.id)
+                successEcho(f"Project \"{project.name}\" successfully selected.")
+
```
dule1322, line 53: Remove newline

coretex/cli/commands/project.py
```diff
@@ -20,47 +20,106 @@
...
+    if type is None:
+        type = project_utils.selectProjectType()
+
+    if description is None:
+        description = clickPrompt("Please enter your project's description:", type = str)
+
+    try:
+        project = Project.createProject(name, ProjectType(type), description)
+
+        if project is not None:
```
dule1322, line 46: `Project.createProject` does not return optional value.

coretex/entities/__init__.py
```diff
@@ -21,4 +21,4 @@
 from .project import *
 from .sample import *
 from .secret import *
-from .task_run import *
+from .task_run import *
```
dule1322, line 24: newline is missing at the end of the file

coretex/entities/project/project.py
```diff
@@ -33,9 +34,16 @@ class Project(BaseObject):
...
+    visibility: ProjectVisibility
 
     @classmethod
-    def createProject(cls, name: str, projectType: ProjectType, description: Optional[str] = None) -> Self:
+    def createProject(
+        cls,
+        name: str,
+        projectType: ProjectType,
+        description: Optional[str] = None,
+        visiblity: Optional[ProjectVisibility] = ProjectVisibility.private
```
dule1322, line 45: Why is `visibility` optional?

coretex/cli/commands/project.py
```diff
@@ -20,47 +20,106 @@
...
-    config["projectId"] = projectId
-    saveConfig(config)
+    defaultProjectId = config.get("projectId")
+    if defaultProjectId is None and name is None:
+        errorEcho(f"To use edit command you need to specifiy project name using \"--project\" or \"-p\" flag, or you can select default project using \"coretex project select\" command.")
+
+    defaultProject = Project.fetchById(config["projectId"])
+
+    if name is None:
+        selectedProject = defaultProject
```
dule1322, line 70: You can assign the value of `defaultProject` variable directly, there is no need for `defaultProject` variable to exist

coretex/cli/commands/project.py
```diff
@@ -20,47 +20,106 @@
...
+    else:
+        selectedProject = Project.fetchOne(name = project)
+
+    if description is None:
+        description = clickPrompt("Please enter new description for your project", type = str, default = selectedProject.description)
+
+    try:
+        selectedProject.update(name = selectedProject.name, description = description)
+        response = networkManager.post("entity-visibility", {
```
dule1322, line 79: Move this to the `Project` class as a separate function for chaning visibility

coretex/cli/commands/project.py
```diff
@@ -20,47 +20,106 @@
...
+            "entity_id": config["projectId"],
+            "type": 1,  # this number represents Project Entity enum value on backend side
```
dule1322, line 81: Can you elaborate on this in more detail?

dule1322 (changes requested):

coretex/cli/commands/project.py
bogdant36: Didn't we want to use only -p or --project flag as part of coretex project flow task? (CTX-5531)

coretex/cli/commands/project.py
bogdant36: So when changing visibility of entity, backend endpoint expects id of entity, type of entity (Project, Dataset, Model etc.) number 1 is representing Project entity on backend side. Do i need to edit comment so its more clear what number 1 represents?

coretex/cli/commands/project.py
dule1322: That is for existing projects, this is creating a new project. Its `coretex project create --project some_name` vs `coretex project create --name some_name`, and the second one feels more intuitive

coretex/cli/commands/project.py
dule1322: That should be converted to an enum if that is the case

coretex/entities/project/project.py
bogdant36: do you agree removing visibility parameter completely? since it does nothing right now, every project will be private no matter what is passed here

coretex/cli/commands/project.py
```diff
@@ -20,23 +20,70 @@
...
+@click.option("--description", "-d", type = str, help = "Project description")
+def create(name: Optional[str], type: Optional[int], description: Optional[str], visibility: Optional[int]) -> None:
+    if name is None:
+        name = ui.clickPrompt("Please enter name of the project you want to create:", type = str)
+
+    if type is None:
+        type = project_utils.selectProjectType()
+
+    if description is None:
+        description = ui.clickPrompt("Please enter your project's description:", type = str)
```
dule1322, line 40: Description is optional. Can user create a project without description using this command?

coretex/cli/commands/project.py
```diff
@@ -20,23 +20,70 @@
...
+
+    try:
+        project = Project.createProject(name, ProjectType(type), description = description)
```
dule1322, line 43: `selectProjectType` should already return a value of type `ProjectType`

coretex/cli/commands/project.py
```diff
@@ -20,23 +20,70 @@
...
+@click.command()
+@click.option("--name", "-n", type = str, help = "Project name")
+@click.option("--type", "-t", type = int, help = "Project type")
+@click.option("--description", "-d", type = str, help = "Project description")
+def create(name: Optional[str], type: Optional[int], description: Optional[str], visibility: Optional[int]) -> None:
+    if name is None:
+        name = ui.clickPrompt("Please enter name of the project you want to create:", type = str)
+
+    if type is None:
+        type = project_utils.selectProjectType()
```
dule1322, line 37: `type` is a reserved name in python, use different name

coretex/cli/commands/project.py
```diff
@@ -20,23 +20,70 @@
...
+            project_utils.selectProject(project.id)
+            ui.successEcho(f"Project \"{project.name}\" successfully selected.")
+    except EntityNotCreated:
+        raise click.ClickException(f"Failed to create project \"{name}\".")
+
+
+@click.command()
+@click.option("--name", "-n", type = str, help = "Project name")
+@click.option("--description", "-d", type = str, help = "Project description")
+def edit(name: Optional[str], description: Optional[str], visibility: Optional[int]) -> None:
```
dule1322, line 57: How do I edit the project name if `name` is used for fetching the project?

coretex/cli/commands/project.py
```diff
@@ -20,23 +20,70 @@
...
+    if defaultProjectId is None and name is None:
+        ui.errorEcho(f"To use edit command you need to specifiy project name using \"--project\" or \"-p\" flag, or you can select default project using \"coretex project select\" command.")
+
+    if name is None and defaultProjectId is not None:
+        selectedProject = Project.fetchById(defaultProjectId)
+    else:
+        selectedProject = Project.fetchOne(name = name)
+
+    if description is None:
+        description = ui.clickPrompt("Please enter new description for your project", type = str, default = selectedProject.description)
```
dule1322, line 69: Descriptions can be quite large so let's not display the default value of the description

coretex/cli/modules/project_utils.py
```diff
@@ -1,10 +1,23 @@
...
 from . import ui
-from ...entities import Project, ProjectType
-from ...networking import EntityNotCreated
+from ...entities import Project, ProjectType, ProjectVisibility
+from ...networking import EntityNotCreated, NetworkResponse, networkManager
+from ...configuration import loadConfig, saveConfig
+
+
+class EntityVisibilityType(IntEnum):
+    project = 1
```
dule1322, line 13: add newline above this line

coretex/cli/modules/project_utils.py
```diff
@@ -23,9 +36,31 @@ def selectProjectType() -> ProjectType:
...
+
+    selectedProjectVisibility = availableProjectVisibilities[selectedChoice]
+    ui.stdEcho(f"You've chosen: {selectedChoice}")
+    return selectedProjectVisibility
+
+
+def changeProjectVisibility(id: int, visibility: ProjectVisibility) -> NetworkResponse:
+    return networkManager.post("entity-visibility", {
+        "entity_id": id,
+        "type": EntityVisibilityType.project,  # this number represents Project Entity enum value on backend side
```
dule1322, line 56: There is no need for the comment anymore

coretex/cli/modules/project_utils.py
```diff
@@ -23,9 +36,31 @@ def selectProjectType() -> ProjectType:
...
+        "visibility": visibility,
+    })
+
+
 def promptProjectCreate(message: str, name: str) -> Optional[Project]:
     if not click.confirm(message, default = True):
-        raise RuntimeError
+        return None
```
dule1322, line 63: What does this affect?

bogdant36: Well in initial implementation for some weird reason I've implemented that If user when prompted to create a new project (if project with specified name is not found) selects No as an answer it will raise RuntimeError, but since there was no description and it was only RuntimeError it would act like return but in the logs you can clearly see the error was raised here, so I've modified this to work as expected. (If user selects no when asked to create new project command will simply stop without any errors in logs and etc.)

coretex/cli/commands/project.py
```diff
@@ -20,23 +20,77 @@
...
+
+        if selectNewProject:
+            project_utils.selectProject(project.id)
+            ui.successEcho(f"Project \"{project.name}\" successfully selected.")
+    except EntityNotCreated:
+        raise click.ClickException(f"Failed to create project \"{name}\".")
+
+
+@click.command()
+@click.option("--project", "-p", type = str, help = "Project name")
```
dule1322, line 57: Change this to argument instead of option

coretex/cli/commands/project.py
```diff
@@ -20,23 +20,77 @@
...
+        raise click.ClickException(f"Failed to edit project \"{selectedProject.name}\".")
 
 
 @click.command()
 @click.option("--name", "-n", type = str, help = "Project name")
-@click.option("--id", type = int, help = "Project ID")
-def select(name: Optional[str], id: Optional[int]) -> None:
-    if name is None and id is None:
-        raise click.UsageError("Please use either --name / -n (for project name) or --id (for project ID)")
+def select(name: Optional[str]) -> None:
```
dule1322, line 91: This should not be optional

coretex/cli/commands/project.py
```diff
@@ -20,23 +20,77 @@
...
+
+        if not response.hasFailed():
+            ui.successEcho(f"Project id \"{config['projectId']}\" successfully edited.")
+
+    except RequestFailedError:
+        raise click.ClickException(f"Failed to edit project \"{selectedProject.name}\".")
 
 
 @click.command()
 @click.option("--name", "-n", type = str, help = "Project name")
```
dule1322, line 90: Can we use argument instead of option here?

coretex/entities/project/project.py
```diff
@@ -33,9 +34,16 @@ class Project(BaseObject):
...
     tasks: List[Task]
+    visibility: ProjectVisibility
 
     @classmethod
-    def createProject(cls, name: str, projectType: ProjectType, description: Optional[str] = None) -> Self:
+    def createProject(
+        cls,
+        name: str,
+        projectType: ProjectType,
+        visiblity: ProjectVisibility = ProjectVisibility.private,
```
dule1322, line 44: Add the function changes to documentation

dule1322 (changes requested):

coretex/cli/commands/project.py
dule1322: this is not done

coretex/entities/project/project.py
```diff
@@ -110,3 +124,10 @@ def addTask(self, name: str, description: Optional[str]) -> bool:
 
         self.tasks.append(task)
         return True
+
+    def changeVisibility(self, visibility: ProjectVisibility) -> NetworkResponse:
```
dule1322, line 128: `updateVisibility`. Function is missing documentation. Raise an exception if the request has failed, do not directly return `NetworkResponse`.

coretex/entities/project/project.py
```diff
@@ -33,16 +35,27 @@ class Project(BaseObject):
...
+    ) -> Self:
         """
             Creates a new project with the provided name and description
 
             Parameters
             ----------
             name : str
                 project name
+            projectType : ProjectType
+                project type of created object
```
dule1322, line 56: Type of the created Project.

coretex/entities/project/project.py
```diff
@@ -33,16 +35,27 @@ class Project(BaseObject):
...
+            visibility : ProjectVisibility
+                visibility type of created object
```
dule1322, line 58: Visibility of the created Project.

dule1322 (changes requested):

coretex/entities/project/project.py
```diff
@@ -110,3 +125,34 @@ def addTask(self, name: str, description: Optional[str]) -> bool:
 
         self.tasks.append(task)
         return True
+
+    def updateVisibility(self, visibility: ProjectVisibility) -> bool:
```
dule1322, line 129: If a function raises an exception if it fails to change visibility, why does it return a bool?

dule1322: Function is updating the visibility in the databse, but locally fetched entity property does not get updated

coretex/entities/project/project.py
```diff
@@ -110,3 +125,30 @@ def addTask(self, name: str, description: Optional[str]) -> bool:
...
+            ----------
+            visibility : ProjectVisibility
+                visibility of the project
+
+            Raises
+            ------
+            NetworkRequestError -> If request for updating the Project visibility failed
+        """
+
+        self.visibility = visibility
```
dule1322, line 143: This should not be updated if request has failed

dule1322 (approved):

PR #166, opened 2024-03-29
CTX-5258: Changed request timeout for sample upload

VukManojlovic:


coretex/networking/network_manager_base.py
```diff
@@ -370,7 +370,8 @@ def formData(
         self,
         endpoint: str,
         params: Optional[Dict[str, Any]] = None,
-        files: Optional[List[FileData]] = None
+        files: Optional[List[FileData]] = None,
+        timeout: TimeoutType = REQUEST_TIMEOUT
```
dule1322, line 374: Add the parameter description into the documentation as well.

PR #179, opened 2024-04-26
CTX-5903: Fixed bug where updated samples don't get unzipped again

VukManojlovic:


coretex/entities/sample/network_sample.py
```diff
@@ -158,6 +159,8 @@ def _download(self, ignoreCache: bool = False) -> None:
 
         if ignoreCache and self.downloadPath.exists():
             self.downloadPath.unlink()
+            if self.path.exists():
```
dule1322, line 162: In case of encrypted Samples `zipPath` should also be deleted

PR #181, opened 2024-04-29
CTX-5758: Always use the same entry point for starting the Node.

bogdant36:
Optimized a CLI flow a bit since there is no more need for double corrections when something in flow changes related to starting the node (only bash template script will be changed nothing more). If actual command for start of the node is called (coretex node start) --force flag is passed to the bash script (since when auto update is performed we don't want that to happen if node is not active, which is not the case here). Rest is the same, formatted code a bit to have more sense.

Resolves CTX-5758.

### DIFF
--- a/coretex/cli/commands/node.py
+++ b/coretex/cli/commands/node.py
@@ -21,9 +21,10 @@ import click
 
 from ..modules import node as node_module
 from ..modules.ui import clickPrompt, stdEcho, successEcho, errorEcho, previewConfig
-from ..modules.update import NodeStatus, getNodeStatus, activateAutoUpdate, dumpScript, UPDATE_SCRIPT_NAME
+from ..modules.update import NodeStatus, getNodeStatus, activateAutoUpdate, dumpScript
 from ..modules.utils import onBeforeCommandExecute
 from ..modules.user import initializeUserSession
+from ..resources import START_SCRIPT_NAME
 from ...configuration import loadConfig, saveConfig, CONFIG_DIR, isNodeConfigured
 from ...utils import docker
 
@@ -54,8 +55,7 @@ def start(image: Optional[str]) -> None:
 
     dockerImage = config["image"]
 
-    if node_module.shouldUpdate(dockerImage):
-        node_module.pull(dockerImage)
+    dumpScript(CONFIG_DIR / START_SCRIPT_NAME, config)
 
     node_module.start(dockerImage, config)
     docker.removeDanglingImages(node_module.getRepoFromImageUrl(dockerImage), node_module.getTagFromImageUrl(dockerImage))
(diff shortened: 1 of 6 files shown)


coretex/cli/resources/update_script_template.sh
```diff
@@ -49,7 +49,13 @@ OUTPUT_DIR="$HOME/.config/coretex"
 mkdir -p "$OUTPUT_DIR"
 
 # Generate the output filename based on the current Unix timestamp
-OUTPUT_FILE="$OUTPUT_DIR/ctx_autoupdate.log"
+if [[ "$1" == "--force" ]]; then
```
dule1322, line 52: So the "--force" flag defines if Node should be started or if automatic update should run? I like this, but lets invest a bit more time into this sh script so that the "coretex node start" only checks if Node is already running and stops it and then runs this sh script with this flag.

dule1322 (changes requested):

coretex/cli/modules/scripts.py
```diff
@@ -73,16 +86,22 @@ def generateUpdateScript(config: Dict[str, Any]) -> str:
     )
 
 
-def dumpScript(updateScriptPath: Path, config: Dict[str, Any]) -> None:
+def dumpScripts(startScriptPath: Path, updateScriptPath: Path, config: Dict[str, Any]) -> None:
```
dule1322, line 89: Replace `startScriptPath` and `updateScriptPath` with a parameter of type `list`

coretex/cli/commands/node.py
```diff
@@ -57,9 +58,11 @@ def start(image: Optional[str]) -> None:
     if node_module.shouldUpdate(dockerImage):
         node_module.pull(dockerImage)
 
-    node_module.start(dockerImage, config)
-    docker.removeDanglingImages(node_module.getRepoFromImageUrl(dockerImage), node_module.getTagFromImageUrl(dockerImage))
+    startScript = (CONFIG_DIR / START_SCRIPT_NAME, generateStartScript(config))
+    updateScript = (CONFIG_DIR / UPDATE_SCRIPT_NAME, generateUpdateScript(config))
+    dumpScripts([startScript, updateScript])
 
+    node_module.start(force = True)
```
dule1322, line 65: `force` is still not removed

coretex/cli/commands/node.py
```diff
@@ -103,9 +106,6 @@ def update() -> None:
         successEcho("Node is already up to date.")
         return
 
-    stdEcho("Updating node...")
-    node_module.pull(dockerImage)
```
dule1322, line 107: This should not be removed, right?

dule1322 (changes requested):

bogdant36:
After conversation with Dusko we decided to implement this on other way and its easier to start from new branch rather than edit this one, closing pull request.

PR #183, opened 2024-04-29
CTX-5645: Implemented entity name validation

VukManojlovic:
- Alphanumeric ("A-Z", "a-z", and "0-9") characters and dash ("-")
- Min length 3
- Max length 50


dule1322:
Were all predefined Tasks updated to generate datasets with this new naming convention?

coretex/entities/utils.py
```diff
@@ -0,0 +1,23 @@
...
+#     GNU Affero General Public License for more details.
+
+#     You should have received a copy of the GNU Affero General Public License
+#     along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+import re
+
+
+def isEntityNameValid(name: str) -> bool:
+    pattern = r"^[A-Za-z0-9-]{3,50}$"
```
dule1322, line 22: Only lowercase alpha-numeric characters should be allowed

coretex/entities/project/task.py
```diff
@@ -71,6 +72,10 @@ def createTask(cls, name: str, projectId: int, description: Optional[str]=None)
                     print("Failed to create task")
         """
 
+        if not isEntityNameValid(name):
+            raise ValueError(">> [Coretex] Task name is invalid. Requirements: alphanumeric characters (\"A-Z\", \"a-z\", and \"0-9\") and dash (\"-\") with length between 3 to 50")
+
+
```
dule1322, line 78: There are 2 newlines here

coretex/entities/dataset/network_dataset.py
```diff
@@ -50,6 +51,9 @@ def _hashDependencies(dependencies: List[str]) -> str:
 
 
 def _chunkSampleImport(sampleType: Type[SampleType], sampleName: str, samplePath: Path, datasetId: int) -> SampleType:
+    if not isEntityNameValid(sampleName):
+        raise ValueError(">> [Coretex] Sample name is invalid. Requirements: alphanumeric characters (\"A-Z\", \"a-z\", and \"0-9\") and dash (\"-\") with length between 3 to 50")
```
dule1322, line 55: Extract the message into a constant since the same one is used everywhere

dule1322 (changes requested):

coretex/entities/dataset/network_dataset.py
```diff
@@ -39,6 +40,7 @@
 
 
 SampleType = TypeVar("SampleType", bound = "NetworkSample")
+NAME_VALIDATION_MESSAGE = ">> [Coretex] Sample name is invalid. Requirements: alphanumeric characters (\"a-z\", and \"0-9\") and dash (\"-\") with length between 3 to 50"
```
dule1322, line 43: This message addresses Sample's only, but is being raised for Dataset as well. Can we update it to say "Entity name is invalid......"

dule1322 (approved):

PR #184, opened 2024-04-30
CTX-5888: Implemented coretex version and update command.

bogdant36:


coretex/cli/modules/utils.py
```diff
@@ -22,6 +22,22 @@
 
 import click
 
+from ...utils.process import command
+
+
+def getVersion() -> Optional[str]:
+    _, output, _ = command(["pip", "show", "coretex"], ignoreStdout = True, ignoreStderr = True)
```
dule1322, line 29: There is a very simple way to do this without parsing command outputs. Let's use that instead of this complex approach.

coretex/cli/modules/utils.py
```diff
@@ -22,6 +22,22 @@
...
+
+    for line in output.split("\n"):
+        if line.startswith("Version:"):
+            return line.split(":", 1)[1].strip()
+
+    return None
+
+
+def updateLib() -> None:
+    command(["pip", "install", "--no-cache-dir", "--upgrade", "coretex"], ignoreStdout = True, ignoreStderr = True)
```
dule1322, line 39: What is user has `pip3` instead of `pip`?

dule1322 (changes requested):

coretex/cli/main.py
```diff
@@ -23,7 +25,27 @@
...
 from .modules.intercept import ClickExceptionInterceptor
+from .modules.utils import updateLib
+
+from ..utils.process import CommandException
+
+
+@click.command()
+def version() -> None:
+        version = getLibraryVersion("coretex")
+        ui.stdEcho(f"Current version of coretex is {version}")
```
dule1322, line 38: Fix indentation

dule1322 (approved):

PR #187, opened 2024-05-07
CTX-5970: Auto-login issue that happens when backend is not a…

bogdant36:
…ble to authorize with saved credentials fix.

### DIFF
--- a/coretex/cli/modules/user.py
+++ b/coretex/cli/modules/user.py
@@ -44,7 +44,7 @@ def authenticateUser(username: str, password: str) -> NetworkResponse:
             raise NetworkRequestError(response, "Something went wrong, please try again later.")
 
         if response.statusCode >= 400:
-            stdEcho(f"Authentification failed with configured credentials. Please try entering your credentials again. Response: {response}")
+            errorEcho(f"Authentification failed with configured credentials. Please try entering your credentials again.")
             authenticate()
 
     return response
--- a/coretex/cli/modules/user.py
+++ b/coretex/cli/modules/user.py
@@ -124,7 +124,7 @@ def initializeUserSession() -> None:
         except NetworkRequestError as ex:
             logging.getLogger("cli").debug(ex, exc_info = ex)
 
-            if 400 <= ex.response.statusCode < 500 :
+            if 400 <= ex.response.statusCode < 500:
                 errorEcho("Authentification failed with configured credentials. Please try entering your credentials again.")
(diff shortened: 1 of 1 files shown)


coretex/cli/modules/user.py
```diff
@@ -44,7 +44,8 @@ def authenticateUser(username: str, password: str) -> NetworkResponse:
             raise NetworkRequestError(response, "Something went wrong, please try again later.")
 
         if response.statusCode >= 400:
-            raise NetworkRequestError(response, "User credentials invalid, please try configuring them again.")
+            stdEcho("Authentification failed with configured credentials. Please try entering your credentials again.")
```
dule1322, line 47: Let's use error output here since the authentication failed.

coretex/cli/modules/user.py
```diff
@@ -97,27 +97,31 @@ def initializeUserSession() -> None:
...
+
+                        if response.statusCode >= 400:
+                            response = authenticateUser(config["username"], config["password"])
+                else:
+                    response = authenticateUser(config["username"], config["password"])
             else:
                 response = authenticateUser(config["username"], config["password"])
-        else:
-            response = authenticateUser(config["username"], config["password"])
+        except:
```
dule1322, line 122: The error should be logged out to the debug logs file

dule1322 (changes requested):

coretex/cli/modules/user.py
```diff
@@ -97,27 +97,31 @@ def initializeUserSession() -> None:
...
+            errorEcho("Authentification failed with configured credentials. Please try entering your credentials again.")
+            authenticate()
```
dule1322, line 124: This should be called only if the request failed with "4xx" error codes

coretex/cli/modules/user.py
```diff
@@ -97,27 +99,34 @@ def initializeUserSession() -> None:
...
+                else:
+                    response = authenticateUser(config["username"], config["password"])
             else:
                 response = authenticateUser(config["username"], config["password"])
-        else:
-            response = authenticateUser(config["username"], config["password"])
+        except NetworkRequestError as ex:
+            logging.getLogger("cli").debug(ex, exc_info = ex)
+
+            if 400 <= ex.response.statusCode < 500 :
```
dule1322, line 127: Fix formatting

dule1322 (approved):

PR #190, opened 2024-05-14
CTX-5959 - CLI - Fix bug where coretex node start fails if ram limit in docker is lower than minimum ram requirement.

bogdant36:


coretex/cli/modules/node.py
```diff
@@ -353,7 +353,7 @@ def configureNode(config: Dict[str, Any], verbose: bool) -> None:
         config["image"] += f":latest-{tag}"
 
     config["storagePath"] = config_defaults.DEFAULT_STORAGE_PATH
-    config["nodeRam"] = int(min(ramLimit, config_defaults.DEFAULT_RAM_MEMORY))
+    config["nodeRam"] = int(min(config_defaults.MINIMUM_RAM_MEMORY, config_defaults.DEFAULT_RAM_MEMORY))
```
dule1322, line 356: The docker ram limit must be used as a default if it is set and higher than the minimum required RAM

dule1322, line 356: RAM stands for "Random Access Memory", there is no need to have a "_MEMORY" suffix in names which have "RAM" as well.

dule1322 (changes requested):

dule1322 (approved):

PR #197, opened 2024-05-27
CTX-6020: Changed generic (network object) .create() function to hand…

bogdant36:
…le network request error and print response instead of having to do that on other places.

### DIFF
--- a/coretex/entities/project/project.py
+++ b/coretex/entities/project/project.py
@@ -18,8 +18,6 @@
 from typing import Optional, Any, List, Dict
 from typing_extensions import Self
 
-from coretex.networking import EntityNotCreated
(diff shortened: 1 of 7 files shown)


coretex/entities/project/project.py
```diff
@@ -82,18 +82,13 @@ def createProject(
...
-        project = cls.create(
+        return cls.create(
             name = name,
             project_task = projectType,
             description = description,
             visiblity = visiblity
         )
 
-        if project is None:
-            raise EntityNotCreated(f"Failed to create project with name \"{name}\"")
```
dule1322, line 93: Let's remove `EntityNotCreated` completely

PR #198, opened 2024-05-28
CTX-5889: Implemented functionality in CLI to check for latest corete…

bogdant36:
…x version and notify user that newer version is available if there is newer version on pypi.


coretex/cli/modules/utils.py
```diff
@@ -17,20 +17,45 @@
...
 
 
+def fetchLatestVersion() -> Optional[str]:
+    url = f"https://pypi.org/pypi/coretex/json"
+    response = requests.get(url)
+    if response.status_code == 200:
+        data = response.json()
+        return str(data['info']['version'])
+    else:
+        return None
```
dule1322, line 44: No need for else. You can use `response.ok` instead of checking `response.status_code` directly

coretex/cli/modules/utils.py
```diff
@@ -17,20 +17,45 @@
...
+from . import ui
 from ...utils.process import command
 
 
 def updateLib() -> None:
     command([sys.executable, "-m", "pip", "install", "--no-cache-dir", "--upgrade", "coretex"], ignoreStdout = True, ignoreStderr = True)
 
 
+def fetchLatestVersion() -> Optional[str]:
+    url = f"https://pypi.org/pypi/coretex/json"
```
dule1322, line 38: String is marked with `f` but no interpolation is done

coretex/cli/modules/utils.py
```diff
@@ -17,20 +17,45 @@
...
+    response = requests.get(url)
+    if response.status_code == 200:
+        data = response.json()
+        return str(data['info']['version'])
```
dule1322, line 42: Use double quotes, not single quotes

dule1322 (changes requested):

coretex/cli/modules/utils.py
```diff
@@ -17,20 +17,45 @@
...
 
+from . import ui
 from ...utils.process import command
 
 
 def updateLib() -> None:
     command([sys.executable, "-m", "pip", "install", "--no-cache-dir", "--upgrade", "coretex"], ignoreStdout = True, ignoreStderr = True)
 
 
+def fetchLatestVersion() -> Optional[str]:
```
dule1322, line 37: This should return a tuple of integers `(major, minor, patch)`.

coretex/cli/modules/utils.py
```diff
@@ -17,20 +17,45 @@
...
+
+
+def checkLibVersion() -> None:
+    current = getLibraryVersion("coretex")
+    latest = fetchLatestVersion()
+
+    if latest is None:
+        return
+
+    if int(current[-3:]) < int(latest[-3:]):
```
dule1322, line 54: Major, minor, and patch versions should all be compared, not just the patch version. Patch version is not guaranteed to be the last 3 digits, what happens if this cannot be converted to integer?

coretex/cli/modules/utils.py
```diff
@@ -17,20 +17,45 @@
...
 def updateLib() -> None:
     command([sys.executable, "-m", "pip", "install", "--no-cache-dir", "--upgrade", "coretex"], ignoreStdout = True, ignoreStderr = True)
 
 
+def fetchLatestVersion() -> Optional[str]:
+    url = "https://pypi.org/pypi/coretex/json"
+    response = requests.get(url)
+    if response.ok:
+        data = response.json()
+        return str(data["info"]["version"])
```
dule1322, line 42: What happens if "info" or "version" are not present anymore?

coretex/cli/modules/utils.py
```diff
@@ -17,20 +17,50 @@
...
+
+def checkLibVersion() -> None:
+    current = getLibraryVersion("coretex")
+    latest = fetchLatestVersion()
+
+    if latest is None:
+        return
+
+    majorCurrent, minorCurrent, patchCurrent = map(int, current.split('.'))
+    majorLatest, minorLatest, patchLatest = map(int, latest.split('.'))
```
dule1322, line 56: Validate both `current` and `latest` strings before splitting and converting them to integers

coretex/cli/modules/utils.py
```diff
@@ -17,20 +17,50 @@
...
     command([sys.executable, "-m", "pip", "install", "--no-cache-dir", "--upgrade", "coretex"], ignoreStdout = True, ignoreStderr = True)
 
 
+def fetchLatestVersion() -> Optional[str]:
+    url = "https://pypi.org/pypi/coretex/json"
+    response = requests.get(url)
+
+    if response.ok:
+        data = response.json()
+        return str(data.get("info", None).get("version", None))
```
dule1322, line 43: `data.get` will return `None` if "info" key is not present which will raise an error. Properly validate that all these values are contained inside of a dictionary and that the are proper types.

coretex/cli/modules/utils.py
bogdant36: Function getLibraryVersion() is returning "string" by default, when I tried to validate this variable I got mypy error that block of code in if statement is unreachable, is there really need to validate "current" variable?

dule1322: There is a need because what if versions has only 2 components, not 3 as we are expecting in our code? What if one of the version components cannot be converted to int?

coretex/cli/modules/utils.py
```diff
@@ -17,20 +17,81 @@
...
+    if response.ok:
+        data = response.json()
+        infoDict = data.get("info")
+
+        if isinstance(infoDict, dict):
+            version = infoDict.get("version")
+            if isinstance(version, str):
+                return version
+
+            logging.debug("Value of json field of key \"version\" in \"info\" dictionary is not of expected type (str).")
```
dule1322, line 51: `logging.getLogger("coretexpylib").debug(">> [Coretex] your-message-goes-here")`

coretex/cli/modules/utils.py
```diff
@@ -17,20 +17,81 @@
...
+    url = "https://pypi.org/pypi/coretex/json"
+    response = requests.get(url)
+
+    if response.ok:
+        data = response.json()
+        infoDict = data.get("info")
+
+        if isinstance(infoDict, dict):
+            version = infoDict.get("version")
+            if isinstance(version, str):
```
dule1322, line 48: Invert conditions to prevent nesting

coretex/cli/modules/utils.py
```diff
@@ -17,20 +17,80 @@
...
 
 def updateLib() -> None:
     command([sys.executable, "-m", "pip", "install", "--no-cache-dir", "--upgrade", "coretex"], ignoreStdout = True, ignoreStderr = True)
 
 
+def fetchLatestVersion() -> Optional[str]:
+    url = "https://pypi.org/pypi/coretex/json"
+    response = requests.get(url)
+
+    if response.ok:
```
dule1322, line 42: Invert this condition

coretex/cli/modules/utils.py
dule1322: This is still not returning this tuple

coretex/cli/modules/utils.py
```diff
@@ -17,20 +17,80 @@
...
+    if not isVersionStrValid(current):
+        logging.getLogger("cli").debug("Current version returned by importlib library is not valid.")
+        return
+
+    if not isVersionStrValid(latest):
+        logging.getLogger("cli").debug("Latest version extracted from pypi website is not valid.")
+        return
+
+    majorCurrent, minorCurrent, patchCurrent = map(int, current.split('.'))
+    majorLatest, minorLatest, patchLatest = map(int, latest.split('.'))
```
dule1322, line 87: What if the int conversion fails?

coretex/cli/modules/ui.py
```diff
@@ -89,6 +89,10 @@ def stdEcho(text: str) -> None:
     click.echo(click.style(f"\n{text}", fg = "cyan"))
 
 
+def warningEcho(text: str) -> None:
+    click.echo(click.style(f"\n WARNING: {text}", fg = "yellow"))
```
dule1322, line 93: Remove space between \n and WARNING

coretex/cli/modules/utils.py
bogdant36: But won't isVersionStrValid() function check if function is valid for conversion, in what case will conversion to int fails if version is validated with isVersionStrValid() function before?

coretex/cli/modules/utils.py
```diff
@@ -17,20 +17,79 @@
...
+    parts = version.split('.')
+
+    if len(parts) != 3:
+        return False
+
+    return all(part.isdigit() for part in parts)
+
+
+def checkLibVersion() -> None:
+    current = getLibraryVersion("coretex")
```
dule1322, line 70: Create a separate function to get the current version which returns a tuple as well

coretex/cli/modules/utils.py
```diff
@@ -15,22 +15,81 @@
...
+    parts = version.split(".")
+
+    if len(parts) != 3:
+        return None
+
+    if all(part.isdigit() for part in parts):
+        major, minor, patch = map(int, version.split('.'))
+        return major, minor, patch
+
+    logging.getLogger("cli").debug(f"Couldn't parse version from string: {version}")
```
dule1322, line 48: Callers of this function should print this log

dule1322 (approved):

PR #211, opened 2024-06-25
CTX-5534: Limit model accuracy to 0-1.

bogdant36:


coretex/entities/model/model.py
```diff
@@ -138,6 +138,8 @@ def createModel(
         if not isEntityNameValid(name):
             raise ValueError(">> [Coretex] Model name is invalid. Requirements: alphanumeric characters (\"a-z\", and \"0-9\") and dash (\"-\") with length between 3 to 50")
 
+        accuracy = max(0, min(accuracy, 1))
```
dule1322, line 141: Print out a warning if accuracy is not a valid value and that it will be clipped to 0-1 range

PR #213, opened 2024-06-25
CTX-5970: Fix for issue when password is changed and authentication h…

bogdant36:
…appens after network request error.


coretex/cli/modules/user.py
```diff
@@ -126,7 +126,10 @@ def initializeUserSession() -> None:
 
             if 400 <= ex.response.statusCode < 500:
                 errorEcho("Authentification failed with configured credentials. Please try entering your credentials again.")
-                authenticate()
+                loginInfo = authenticate()
+                config = saveLoginData(loginInfo, config)
+                saveConfig(config)
+                return
```
dule1322: Let's redesign the flow for this function. We've already had multiple issues with it so let's rewrite it to be more straightforward

PR #222, opened 2024-06-27
CTX-6118 Since this can be potential cause of configuration resetting…

bogdant36:
… added debug log to track deeper if maybe config file becomes corrupted and it resets because json.load() function is not able to read the config file.

### DIFF
--- a/coretex/configuration.py
+++ b/coretex/configuration.py
@@ -21,6 +21,7 @@ from pathlib import Path
 import os
 import json
 import sys
+import logging
 
 
 def isCliRuntime() -> bool:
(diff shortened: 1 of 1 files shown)


coretex/configuration.py
```diff
@@ -62,6 +63,7 @@ def loadConfig() -> Dict[str, Any]:
         try:
             config: Dict[str, Any] = json.load(configFile)
         except json.JSONDecodeError:
+            logging.getLogger("cli").debug(f"Failed to load corrupted config file. Content: {configFile.read()}")
```
dule1322, line 66: Log out the exception as well with a full stack trace. Also `json.load` will read this file as well which will cause `configFile.read` to raise an error the 2nd time it is called. The file should only be read once

bogdant36: So I can just use configFile as if I am calling configFile.read() after initial .read() is called?

PR #223, opened 2024-06-28
CTX-5921: Swap limit check implemented.

bogdant36:


coretex/cli/modules/node.py
```diff
@@ -255,6 +255,23 @@ def promptRam(config: Dict[str, Any], ramLimit: int) -> int:
     return nodeRam
 
 
+def promptSwap(config: Dict[str, Any], swapLimit: int) -> int:
+    nodeSwap: int = clickPrompt(
+        f"Node SWAP memory limit in GB (Maximum: {swapLimit}GB) (press enter to use default)",
+        swapLimit,
```
dule1322, line 261: Default should be `swapLimit` or configured RAM value x2, whichever is lower.

coretex/cli/modules/node.py
```diff
@@ -355,8 +377,8 @@ def configureNode(config: Dict[str, Any], verbose: bool) -> None:
 
         config["cpuCount"] = promptCpu(config, cpuLimit)
         config["nodeRam"] = promptRam(config, ramLimit)
+        config["nodeRam"] = promptSwap(config, swapLimit)
```
dule1322, line 380: `nodeRam` is being set two times here

coretex/utils/docker.py
```diff
@@ -147,3 +150,30 @@ def getResourceLimits() -> Tuple[int, int]:
...
+        return None
+
+
+def getDockerSwapLimit() -> int:
+    configPath = getDockerConfigPath()
+
+    if configPath is None:
+        return getTotalSwapMemory()
+
+    with configPath.open("r") as configFile:
```
dule1322, line 172: Where do you check if the `configPath` exists?

coretex/utils/docker.py
```diff
@@ -147,3 +150,30 @@ def getResourceLimits() -> Tuple[int, int]:
...
+        configJson = json.load(configFile)
+
+    swapLimit = configJson.get("swapMiB")
+    if not isinstance(swapLimit, int):
```
dule1322, line 176: if `swapLimit` is not valid do not raise an error, return either 0, or if it will work return the value of swap configured on the system.

coretex/cli/modules/node.py
```diff
@@ -255,6 +255,23 @@ def promptRam(config: Dict[str, Any], ramLimit: int) -> int:
...
+    nodeSwap: int = clickPrompt(
+        f"Node SWAP memory limit in GB (Maximum: {swapLimit}GB) (press enter to use default)",
+        swapLimit,
+        type = int
+    )
+
+    if nodeSwap > swapLimit:
+        errorEcho(
+            f"ERROR: SWAP memory limit in Docker Desktop ({swapLimit}GB) is lower than the configured value ({nodeSwap}GB). "
+            "Please adjust resource limitations in Docker Desktop settings."
```
dule1322, line 268: Is there a need for the 2nd sentence? You can tell user that if they want to use a **higher value** than what is currently configured that they also need to change docker limits, not docker desktop limits.

dule1322 (changes requested):

coretex/cli/modules/node.py
```diff
@@ -255,6 +255,23 @@ def promptRam(config: Dict[str, Any], ramLimit: int) -> int:
     return nodeRam
 
 
+def promptSwap(config: Dict[str, Any], swapLimit: int) -> int:
```
dule1322, line 258: `config` is only used to extract ram value, why not pass in just that, instead of whole config object?

coretex/cli/modules/node.py
```diff
@@ -355,8 +377,8 @@ def configureNode(config: Dict[str, Any], verbose: bool) -> None:
 
         config["cpuCount"] = promptCpu(config, cpuLimit)
         config["nodeRam"] = promptRam(config, ramLimit)
+        config["nodeSwap"] = promptSwap(config.get("nodeRam", 0), swapLimit)
```
dule1322, line 380: nodeRam must exist at this point, it is previously configured with `promptRam` function

dule1322 (approved):

PR #224, opened 2024-06-28
CTX-5921: Added new condition in onBeforeCommandExecute() callback wh…

bogdant36:
…ich will check if "command_name" is inside excludeOptions also.

### DIFF
--- a/coretex/cli/main.py
+++ b/coretex/cli/main.py
@@ -61,7 +61,7 @@ def update() -> None:
 
 
 @click.group(cls = ClickExceptionInterceptor)
-@utils.onBeforeCommandExecute(utils.checkLibVersion)
+@utils.onBeforeCommandExecute(utils.checkLibVersion, excludeOptions = ["update"])
 def cli() -> None:
     pass
 
--- a/coretex/cli/main.py
+++ b/coretex/cli/main.py
@@ -61,7 +61,7 @@ def update() -> None:
 
 
 @click.group(cls = ClickExceptionInterceptor)
-@utils.onBeforeCommandExecute(utils.checkLibVersion, excludeOptions = ["update"])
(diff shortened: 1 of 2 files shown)


coretex/cli/main.py
```diff
@@ -61,7 +61,7 @@ def update() -> None:
 
 
 @click.group(cls = ClickExceptionInterceptor)
-@utils.onBeforeCommandExecute(utils.checkLibVersion)
+@utils.onBeforeCommandExecute(utils.checkLibVersion, excludeOptions = ["update"])
```
dule1322, line 64: Will this exclude `coretex node update` as well, or just `coretex update` command?

coretex/cli/modules/utils.py
```diff
@@ -157,6 +157,9 @@ def onBeforeCommandExecute(fun: Callable[..., Any], excludeOptions: Optional[Lis
     def decorator(f: Callable[..., Any]) -> Callable[..., Any]:
         @wraps(f)
         def wrapper(*args: Any, **kwargs: Any) -> Any:
+            if click.get_current_context().invoked_subcommand in excludeOptions:
```
dule1322, line 160: `excludedOptions` is used for excluding parametar of type option when an command is run. Can we add `excludedCommands`, or `excludedSubcommands` parameter to use in cases like these?

dule1322 (changes requested):

coretex/cli/main.py
bogdant36: just coretex update command.

dule1322: This is still using `excludeOptions`, shouldn't it use `excludeSubcommands`?

dule1322 (approved):

PR #233, opened 2024-07-10
CTX-5524: Created function for running inference on ONNX models with the option to perform verification

VukManojlovic:


coretex/utils/inference.py
```diff
@@ -0,0 +1,81 @@
...
+from pathlib import Path
+
+import json
+import uuid
+import asyncio
+
+from onnxruntime import InferenceSession
+
+import ezkl
+import numpy as np
```
bogdant36, line 11: sort imports correctly

VukManojlovic: I'm not sure what you are requesting. Can you please elaborate?

PR #236, opened 2024-07-15
CTX-5524: Added tag management and inference.py hotfix

VukManojlovic:


coretex/entities/tag.py
```diff
@@ -0,0 +1,144 @@
...
+            "name": tagName,
+            "type": int(self.entityTagType),
+            "project_id": self.projectId,
+        }
+
+        response = networkManager.get("tag", parameters)
+        if response.hasFailed():
+            raise NetworkRequestError(response, "Failed to check existing tags")
+
+        tags = response.getJson(dict)["data"]
```
dule1322, line 56: Check if `data` field exists and is of correct type when accessing it.

coretex/entities/tag.py
```diff
@@ -0,0 +1,144 @@
...
+        if len(tags) == 0:
+            return None
+
+        tagId: int = tags[0]["id"]
```
dule1322, line 60: Check if `id` field exists and is of correct type when accessing it.

coretex/entities/tag.py
```diff
@@ -0,0 +1,144 @@
...
+    @property
+    @abstractmethod
+    def entityTagType(self) -> int:
+        pass
+
+    def _getTagId(self, tagName: str) -> Optional[int]:
+        parameters = {
+            "name": tagName,
+            "type": int(self.entityTagType),
+            "project_id": self.projectId,
```
dule1322, line 49: Remove `,` from the last member of dict

dule1322, line 49: Why is `project_id` needed to fetch tags?

dule1322 (changes requested):

coretex/entities/tag.py
VukManojlovic: This way only tags of the relevant project are returned. We then check if the inputted tag already exists for that project.

coretex/entities/tag.py
```diff
@@ -0,0 +1,152 @@
...
+
+    @property
+    @abstractmethod
+    def entityTagType(self) -> int:
+        pass
+
+    def _getTagId(self, tagName: str) -> Optional[int]:
+        parameters = {
+            "name": tagName,
+            "type": int(self.entityTagType),
```
bogdant36, line 48: isn't entityTagType already of type int? why is casting to int necessary here?

coretex/entities/tag.py
```diff
@@ -0,0 +1,152 @@
...
+            "tags": tags
+        }
+
+        response = networkManager.post("tag/entity", parameters)
+        if response.hasFailed():
+            raise NetworkRequestError(response, "Failed to create tag")
+
+    def removeTag(self, tag: str) -> None:
+        """
+            Remove tag with provided name from the model
```
bogdant36, line 126: does this function removes tag from model entity only?

coretex/utils/inference.py
bogdant36, line 1: Why are there changes to inference.py file that are not related to tag management like described in pr title and description. This makes changes hard to track. As far as I've seen CTX-5524 is related to Endpoints - Zero-knowledge ML, I don't see any tag related comments that should be done in scope of this task.

coretex/entities/tag.py
VukManojlovic: Backend receives a EntityTagType class for the get request (which doesn't happen for post requests) so I cast it to int even though it behaves like an integer object in python code.

coretex/utils/inference.py
VukManojlovic: CTX-5524 is the first task that requests tags to be added programmatically, so the coretexpylib code for that was also included as part of this task. I will add a comment that will clarify this.

inference.py was not initially included in the PR, and the change done to it is a minor hotfix for verified inference, so the title never changed. I see now that the title may be confusing, so I will make it more accurate for the PR.

coretex/entities/tag.py
dule1322: Why not just change the return type of `entityTagType` to `EntityTagType` and then instead of casting it to `int` here you can do it like `self.entityTagType.value`.

VukManojlovic: The exact value return value of `EntityTagType.dataset` is `Literal[EntityTagType.dataset]` and similarly `.model` for model. How about we instead keep `int` as the return type of `entityTagType` and return EntityTagType.dataset.value / .model.value.
This way we keep a standardized return value for `entityTagType`. All uses of this property need the enum integer value anyway.

dule1322: But that would allow returning any integer, it's better to use enums to limit the number of available values which can be returned to prevent potential errors during coding. If you put `int` as the return type you can return any kind of integer for that, whilst if the return type of the property is `EntityTagType` mypy will complain if you do not return a valid value from that enum.

dule1322 (approved):

PR #237, opened 2024-07-16
CTX-5517: Implement Project Secret type

dule1322:


coretex/entities/project/project.py
```diff
@@ -36,6 +36,8 @@ class Project(BaseObject):
 
     tasks: List[Task]
     visibility: ProjectVisibility
+    isEncrypted: bool
+    secretName: Optional[str]
```
igorperic17, line 40: Wouldn't it be better if isEncrypted was a computed property? 

`isEncrypted = secretName is not None`

Also, do we want only a secret name or a full `ProjectSecret` object here?

igorperic17 (changes requested):

coretex/entities/project/project.py
dule1322: This is a model from backend which contains these fields so both `secretName` and `isEncrypted` come from the backend model.

`ProjectSecret` cannot be fetched here as this `secretName` value can be used for fetching tow different versions of the secret - user secret (project key encrypted with user's public key) or Node secret (project key encrypted with Node's public key).

igorperic17: Perhaps backend can be fixed as well then to remove a redundant boolean.

igorperic17: Regarding the second comment, couldn't the logic for decryption be encapsulated in some interface or parent class, like `UserSecret` or `NodeSecret`?

dule1322: The logic for decryption is the same the only thing that is different is the key, but the same method is used for decryption.

The fetching of the secrets is different - `SecretType.fetchByName`, and `SecretType.fetchNodeSecret`. Now what you want to use for fetching depends on what your goal is? If the Node is running a Workflow in an encrypted Project it will use `SecretType.fetchNodeSecret` to get the decryption key, whilst if CLI is being used to run local Workflow `SecretType.fetchByName` will be used.

PR #242, opened 2024-08-05
CTX-6271: Implemented deactivate node functionality so there is no mo…

bogdant36:
…re waiting for reconnection in 60s window when node is turned off and started again.

### DIFF
--- a/coretex/cli/commands/node.py
+++ b/coretex/cli/commands/node.py
@@ -34,6 +34,11 @@ from ...utils import docker
 def start(image: Optional[str]) -> None:
     config = loadConfig()
 
+    if config.get("nodeId") is None:
+        config["nodeId"] = node_module.fetchNodeId(config["nodeName"])
+
+    saveConfig(config)
+
     if node_module.isRunning():
         if not clickPrompt(
             "Node is already running. Do you wish to restart the Node? (Y/n)",
@@ -71,6 +76,10 @@ def stop() -> None:
         errorEcho("Node is already offline.")
         return
 
+    if config.get("nodeId") is None:
+        config["nodeId"] = node_module.fetchNodeId(config["nodeName"])
+
+    saveConfig(config)
     node_module.stop(config["nodeId"])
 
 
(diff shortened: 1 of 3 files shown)


coretex/configuration.py
```diff
@@ -122,7 +122,8 @@ def isNodeConfigured(config: Dict[str, Any]) -> bool:
         config.get("nodeRam") is not None and
         config.get("nodeSwap") is not None and
         config.get("nodeSharedMemory") is not None and
-        config.get("nodeMode") is not None
+        config.get("nodeMode") is not None and
+        config.get("nodeId") is not None
```
dule1322, line 126: Is this change backwards compatible?

coretex/cli/modules/node.py
```diff
@@ -458,7 +472,7 @@ def initializeNodeConfiguration() -> None:
             errorEcho("If you wish to reconfigure your node, use \"coretex node stop\" command first.")
             return
 
-        stop()
+        stop(config["nodeId"])
```
dule1322, line 475: `nodeId` which is saved in `config` is of `str` type (that is what `registerNode` functions returns) but `stop` expects it to be `int`. Change `registerNode` to return a `int` properly.

dule1322 (changes requested):

coretex/cli/modules/node.py
```diff
@@ -117,10 +117,21 @@ def clean() -> None:
         raise NodeException("Failed to clean inactive Coretex Node.")
 
 
-def stop() -> None:
+def deactivateNode(id: int) -> None:
+    params: Dict[str, Any] = {
+        "id": id,
```
dule1322, line 122: Remove comma

coretex/cli/modules/node.py
```diff
@@ -117,10 +117,21 @@ def clean() -> None:
         raise NodeException("Failed to clean inactive Coretex Node.")
 
 
-def stop() -> None:
+def deactivateNode(id: int) -> None:
+    params: Dict[str, Any] = {
```
dule1322, line 121: There is no need for explicit type here

VukManojlovic (approved):

coretex/cli/commands/node.py
```diff
@@ -41,13 +43,11 @@ def start(image: Optional[str]) -> None:
         ):
             return
 
-        node_module.stop()
+        node_module.stop(config["nodeId"])
```
dule1322, line 46: This will raise an error if `nodeId` does not exist in configuration. Use Node name to fetch the Node id if it isn't saved in the config.

dule1322 (approved):

PR #254, opened 2024-08-20
CTX-6562: Implement coretex node status command.

bogdant36:


coretex/cli/commands/node.py
```diff
@@ -179,6 +179,11 @@ def config(advanced: bool) -> None:
     activateAutoUpdate()
 
 
+@click.command()
+def status() -> None:
+    ui.stdEcho(f"Current status of Node is {getNodeStatus().name}.")
```
dule1322, line 184: Can you color the status based on what it is?
- inactive: red
- active: green
- busy: blue
- reconnecting: yellow

PR #258, opened 2024-08-23
CTX-6577(part): serverUrl bug fix and added prompt for specifying ser…

bogdant36:
…ver url rather than having to override env var manually.


coretex/cli/modules/user.py
```diff
@@ -15,16 +15,44 @@
...
+def configureServerUrl() -> str:
+    availableChoices = ["Official Coretex Server URL", "Custom Server URL"]
+    selectedChoice = ui.arrowPrompt(availableChoices, "Please select server url that you want to use (use arrow keys to select an option):")
+
+    if not "Official" in selectedChoice:
+        serverUrl: str = ui.clickPrompt("Enter server url that you wish to use", type = str)
+
+        if not validateServerUrl(serverUrl):
+            ui.warningEcho("You've entered invalid server url. Please try again.")
+            serverUrl = configureServerUrl()
```
dule1322, line 43: It should display only a `ui.clickPrompt`, not the choice again

dule1322 (changes requested):

coretex/cli/modules/user.py
```diff
@@ -15,16 +15,43 @@
...
+from ...networking import networkManager, NetworkRequestError, baseUrl
 from ...configuration import UserConfiguration, InvalidConfiguration, ConfigurationNotFound
 
 
+def validateServerUrl(serverUrl: str) -> bool:
+    try:
+        endpoint = baseUrl(serverUrl) + "/info.json"
+        response = requests.get(endpoint, timeout = 5)
+        return response.ok
+    except BaseException as ex:
```
dule1322, line 30: `BaseException` should never be catched. In this case using `requests.exceptions.RequestException` is more fitting.

dule1322 (changes requested):

coretex/cli/modules/user.py
```diff
@@ -15,16 +15,43 @@
...
+
+def configureServerUrl() -> str:
+    availableChoices = ["Official Coretex Server URL", "Custom Server URL"]
+    selectedChoice = ui.arrowPrompt(availableChoices, "Please select server url that you want to use (use arrow keys to select an option):")
+
+    if not "Official" in selectedChoice:
+        serverUrl: str = ui.clickPrompt("Enter server url that you wish to use", type = str)
+
+        if not validateServerUrl(serverUrl):
+            serverUrl = ui.clickPrompt("You've entered invalid server url. Please try again.", type = str)
```
dule1322, line 42: This will only prompt once, it should prompt until the user has entered valid url

PR #263, opened 2024-09-02
CTX-6321: Implementation of coretex node --version command.

bogdant36:
CTX-6321: Implementation of coretex node --version command regarding new changes that we made to our Docker image (storing "version" property of image in Config["Labels"] dictionary).

### DIFF
--- a/coretex/cli/modules/node.py
+++ b/coretex/cli/modules/node.py
@@ -200,8 +200,8 @@ def shouldUpdate(image: str) -> bool:
     return True
 
 
-def getNodeVersion(image: Optional[str] = None) -> str:
-    if image is None:
+def getNodeVersion() -> str:
+    if not isRunning():
         try:
             nodeConfig = NodeConfiguration.load()
             image = nodeConfig.image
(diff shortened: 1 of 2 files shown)


coretex/cli/modules/node.py
```diff
@@ -200,6 +200,35 @@ def shouldUpdate(image: str) -> bool:
     return True
 
 
+def getNodeVersion(image: Optional[str] = None) -> str:
```
igorperic17, line 203: Image should be removed as param from here. It should be read from the running container (if any) and taken from config as a failback.

igorperic17 (changes requested):

coretex/cli/modules/node.py
```diff
@@ -200,6 +200,37 @@ def shouldUpdate(image: str) -> bool:
     return True
 
 
+def getNodeVersion() -> str:
+    if not isRunning():
+        try:
+            nodeConfig = NodeConfiguration.load()
+            image = nodeConfig.image
+        except ConfigurationNotFound:
```
igorperic17, line 208: we need dev ergonomic typing support for exceptions here ASAP

igorperic17: @dule1322

